PR #92, opened 2026-05-06
fix(ci): pin cosign to v2.6.0 to restore sign-blob output

### LINKED ISSUE #93: bug: cosign v3.x breaks sign-blob in release workflow
## Summary

The release workflow fails at the "Sign SHA256SUMS with Cosign (keyless)" step with:

```
Error: signing SHA256SUMS.txt: create bundle file: open : no such file or directory
```

## Root Cause

`sigstore/cosign-installer@v4.1.1` without a pinned version installs the latest cosign — currently v3.x. Cosign v3.x changed `sign-blob` to use `--new-bundle-format` by default, which:

- silently ignores `--output-certificate` and `--output-signature`
- emits a single `.bundle` file instead
- requires `--bundle <path>` to be passed explicitly

The workflow passes neither `--bundle` nor pins the cosign version, so the bundle path is empty and the file-open fails.

## Impact

Every skill release using this reusable workflow is broken. No new skill versions can be published until this is resolved.

## Reproduction

1. Push an annotated signed tag to any skill repo that uses this reusable workflow
2. The "Sign SHA256SUMS with Cosign (keyless)" step fails

## Fix Options

**Short-term (PR #92):** Pin `cosign-release: v2.6.0` — restores the two-file output without any other changes needed.

**Long-term:** Migrate to the cosign v3 bundle format:
- Replace `--output-certificate` / `--output-signature` with `--bundle SHA256SUMS.txt.bundle`
- Update `cosign verify-blob` to use `--bundle`
- Update release artifact attachments (`.bundle` instead of `.pem` + `.sig`)
- Update consumer documentation

## Related

- Fix PR: #92

Gitsko:
Closes #93

## Problem

`cosign sign-blob` fails in the release workflow with:

```
Error: signing SHA256SUMS.txt: create bundle file: open : no such file or directory
```

Cosign v3.x (installed via `sigstore/cosign-installer@v4.1.1` without a pinned version) changed `sign-blob` to emit a **single bundle file** by default (`--new-bundle-format`). This silently ignores `--output-certificate` and `--output-signature`, and instead expects `--bundle <path>`. Since neither flag is passed, the bundle path is empty → the file-open fails.

## Fix

Pin `cosign-release: v2.6.0` in the installer step. In v2.x the two-file output (`SHA256SUMS.txt.pem` + `SHA256SUMS.txt.sig`) is still the default, so the existing sign and verify steps work without any other changes.

## Why a workaround instead of migrating to v3 bundle format?

Migrating to the bundle format would also require updating:
- the `cosign verify-blob` step (replace `--certificate` + `--signature` with `--bundle`)
- the `Create GitHub Release` step (attach `.bundle` instead of `.pem`/`.sig`)
- documentation in skill READMEs and any consumer tooling

That is a larger change worth a dedicated PR. This fix unblocks all skill releases immediately.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -237,6 +237,15 @@ jobs:
 
       - name: Install Cosign
         uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003 # v4.1.1
+        with:
+          # Pin to cosign v2.x. Cosign v3.x switched `sign-blob` to
+          # `--new-bundle-format` by default, which deprecates
+          # `--output-certificate` / `--output-signature` and instead
+          # requires `--bundle <path>`. Until this workflow is migrated
+          # to emit a single bundle file (and the verify step + caller
+          # release artifacts updated to match), stay on v2.x where
+          # the legacy two-file output is still the default.
+          cosign-release: v2.6.0
 
       - name: Sign SHA256SUMS with Cosign (keyless)
         working-directory: dist/releases


CybotTM:
Closing in favor of [#94](https://github.com/netresearch/skill-repo-skill/pull/94), which has been merged (commit [cf898d4](https://github.com/netresearch/skill-repo-skill/commit/cf898d4)) and resolves [#93](https://github.com/netresearch/skill-repo-skill/issues/93).

To be clear, the change in this PR was correct as a triage fix — pinning `cosign-release: v2.6.0` would have unblocked releases immediately with a 9-line diff and zero consumer-facing breakage. Nothing wrong with the diagnosis or the patch.

The reason we went with #94 instead is that the v2.x pin would have left a deferred migration on the table:

- **cosign v2.x is on the deprecation track.** Sigstore is pushing the bundle format as the canonical output of `sign-blob`. Staying on v2 means tracking its CVEs and revisiting this exact piece of the workflow later anyway.
- **Those follow-up changes were going to be required regardless.** Whichever path we took today, we'd eventually need to update:
  - `release.yml` sign step → `--bundle SHA256SUMS.txt.bundle`
  - `release.yml` verify step → `--bundle` instead of `--certificate` + `--signature`
  - `release.yml` header comment + inline verify recipe
  - `skills/skill-repo/references/release-discipline.md` consumer recipe
- **The migration turned out to be small** — 22 lines including the explanatory comments, all in this one repo (no other consumer-facing surfaces in the org reference `.pem`/`.sig` directly). Doing it now skips a future PR.

So the tradeoff was: small immediate diff + future cleanup PR (this PR), vs. slightly larger diff now + done (#94). With the migration being this small, the second option won.

Trade-off is acknowledged in #94's description for the record. Thanks for the quick triage and root-cause writeup in #93 — that's what made the v3 path obvious.

PR #123, opened 2026-06-16
feat(trigger+SR-37): add release keywords to trigger; add version-parity checkpoint

aseemann:
## Summary

During a skill release session the `skill-repo` skill never auto-triggered because \"release\", \"tag\", and \"version bump\" were absent from the trigger description. The result: `SKILL.md` and `composer.json` were bumped to `v1.1.0` but `.claude-plugin/plugin.json` was missed, causing CI to fail with:

```
SKILL.md metadata.version '1.1.0' != plugin.json version '1.0.0'
```

The knowledge to prevent this already existed (`check-version-parity.sh`, `release-discipline.md`) — the skill just didn't fire.

## Changes

**`SKILL.md` trigger** — appended release-time keywords so the skill fires when an agent starts a version bump or tag workflow:
> _…or releasing a new skill version (version bump, tagging, plugin.json / SKILL.md / composer.json version parity, git tag push)._

**`checkpoints.yaml` SR-37** — new mechanical checkpoint that runs `check-version-parity.sh` and fails if any `[ERROR]` line appears. Catches the mismatch as a reviewable finding before the tag is pushed, even if the skill was never explicitly invoked.

## Test plan

- [ ] CI passes on this PR
- [ ] Verify SR-37 fires on a repo where `plugin.json` version ≠ `SKILL.md` version
- [ ] Verify SR-37 passes on a repo where versions are in sync

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## Revisions while finishing (fb783b6)

- Rebased onto main (was BEHIND).
- `ddf429c` trims the trigger wording: the keyword addition pushed SKILL.md to 513 words, over the 500-word CI cap. All release keywords survive (releasing, version bump, tagging, version-parity); now 498 words.
- `fb783b6` replaces the SR-37 pattern. The original could never fire: the parity script prints `ERROR:`, never `[ERROR]`, so the negated grep always passed — and the assessment runner's command allowlist rejects `bash` as base command, so the runner would have reported "Command rejected" permanently. New pattern is a single-line awk (allowlist-compliant: no `;`, no `&&`) comparing plugin.json `.version` against every `metadata.version` in `skills/*/SKILL.md` frontmatter, with a frontmatter gate so JSON examples in SKILL.md bodies do not match. Verified: parity rc=0, either-file-bumped rc=1, gawk+mawk, `is_safe_eval_command` accepts it.

### DIFF
--- a/skills/skill-repo/SKILL.md
+++ b/skills/skill-repo/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: skill-repo
-description: "Use when creating new skill repositories from scratch, standardizing or validating existing skill repo structure, setting up composer/release workflows for skills, configuring split licensing (MIT + CC-BY-SA-4.0), fixing plugin.json / SKILL.md validation errors, or releasing a new skill version (version bump, tagging, plugin.json / SKILL.md / composer.json version parity, git tag push)."
+description: "Use when creating skill repositories, standardizing or validating skill repo structure, setting up composer/release workflows, configuring split licensing (MIT + CC-BY-SA-4.0), fixing plugin.json / SKILL.md validation or version-parity errors, or releasing a skill version (version bump, tagging)."
 license: "(MIT AND CC-BY-SA-4.0). See LICENSE-MIT and LICENSE-CC-BY-SA-4.0"
 compatibility: "Requires bash 4.3+, python3."
 metadata:
--- a/skills/skill-repo/SKILL.md
+++ b/skills/skill-repo/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: skill-repo
-description: "Use when creating new skill repositories from scratch, standardizing or validating existing skill repo structure, setting up composer/release workflows for skills, configuring split licensing (MIT + CC-BY-SA-4.0), or fixing plugin.json / SKILL.md validation errors."
+description: "Use when creating new skill repositories from scratch, standardizing or validating existing skill repo structure, setting up composer/release workflows for skills, configuring split licensing (MIT + CC-BY-SA-4.0), fixing plugin.json / SKILL.md validation errors, or releasing a new skill version (version bump, tagging, plugin.json / SKILL.md / composer.json version parity, git tag push)."
 license: "(MIT AND CC-BY-SA-4.0). See LICENSE-MIT and LICENSE-CC-BY-SA-4.0"
 compatibility: "Requires bash 4.3+, python3."
 metadata:
--- a/skills/skill-repo/checkpoints.yaml
+++ b/skills/skill-repo/checkpoints.yaml
@@ -253,6 +253,20 @@ mechanical:
     severity: warning
     desc: "Release workflow should not declare workflow_dispatch or use --admin merges — the auto-bump path produced unsigned tags. Use signed tag-push only (see skill-repo-skill PR #15)."
 
+  - id: SR-37
+    type: command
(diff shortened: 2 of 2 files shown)


CybotTM:
@aseemann I cannot confirm the issue you are describing. Do you miss the github-release skill which is responsible for release handling.

Also it seems you are missing pre-commit checks on your side. The lengthy SKILL.md should ahve been blocked locally already.

CybotTM:
Thanks @aseemann — a few things to sort out before this goes further.

I first read this against *this* repo's release conventions, but you were almost certainly releasing a **different** org skill repo (this PR targets `skill-repo` because it's the meta-skill *about* releasing skill repos). Under that reading the incident may well be real — so the question is what the right fix is and where it belongs.

### 1. Which repo hit the mismatch?

If you can point me at the repo whose release failed, I can check its `validate.yml` / pre-commit setup. The likely root cause: that repo **hadn't adopted skill-repo-skill's standard template + hooks**, so it had no PR-time and no local parity guard, and the mismatch only surfaced at tag/release time. If so, the durable fix is getting that repo onto the standard template + pre-commit hooks (so CI enforces parity everywhere), not only broadening this skill's trigger — which only helps `skill-repo-skill` itself unless it propagates.

### 2. Releases are already owned by the `github-release` skill

That skill triggers on "release / version bump / tag" (and even blocks raw `gh release`). So broadening `skill-repo`'s trigger overlaps the skill that already owns this flow — worth deciding whether this discipline belongs in `github-release` rather than here.

### 3. This PR's own CI is red — and that's a signal in itself

`Skill Validation` fails because the broadened description pushes `SKILL.md` to **510 words**, over the hard **500-word cap** in `validate-skill.sh` (main is 492). This repo ships a `.pre-commit-config.yaml` with the `validate-skill` and `check-version-parity` hooks — so this should have been **blocked at commit time locally**, never reaching a red PR. That it didn't suggests the pre-commit hooks aren't installed in your agent's dev environment (or were bypassed with `--no-verify`) — the same class of gap that may have let the original release through.

### Questions

1. **Which repo's release hit the parity mismatch?**
2. **Was this PR generated by `/retro:retro`?** If so I'd want to check how it reconstructed the incident.
3. Are this repo's pre-commit hooks installed in your agent environment? (`pre-commit install` / `npm install`)

### Prompt you can paste into your agent to diagnose the env gap

```
This repo ships a .pre-commit-config.yaml that runs validate-skill and
check-version-parity, and these same checks run in CI. A PR I authored
reached GitHub with a red "Skill Validation" check (SKILL.md over the
500-word cap) — the local pre-commit hook should have blocked that commit.

Investigate why the local hook did not fire:
1. Is pre-commit installed?  `command -v pre-commit && pre-commit --version`
2. Are the git hooks installed in this repo?  `ls -la .git/hooks/pre-commit`
   and check `git config --get core.hooksPath`
3. Run the hooks manually against my change:
   `pre-commit run --all-files`  and  `pre-commit run validate-skill --all-files`
4. If pre-commit isn't installed, install it and wire it up:
   `pre-commit install --install-hooks`
5. Confirm the hook now catches the 500-word overflow before commit.

Report: was the hook missing, uninstalled, or bypassed (--no-verify)? Then
fix the environment so CI-equivalent checks run locally before every commit.
```

Once I know the repo I can recommend whether to keep, retarget, or close this. The 510-word cap break needs fixing regardless.

PR #211, opened 2026-08-10
feat(quality): run the tests, and see the checks nobody runs

CybotTM:
`repository-quality-rules.md` has said since it was written that a mechanically checkable requirement shipped as prose is a **FAIL**. Nothing enforced it, and a fleet measurement across `~/p` shows the cost: **33 skill repos ship scripts, 27 of them have no test file at all — 276 scripts with nothing exercising them.**

The cause is structural rather than sloppiness: this repo offers 20 reusable workflows and **not one of them runs `tests/`**. A test file in a consumer repo executes only if that repo hand-rolls its own workflow, so almost none do. In `git-workflow` that let two defects sit in `verify-git-workflow.sh` indefinitely — `set -e` with `((WARNINGS++))` aborted the script after the first warning (11 of 12 sections never ran), and `[[ ! -d ".git" ]]` refused to run in any worktree. Both were found by writing the first test that repo's verifier ever had.

## What this adds

**`.github/workflows/tests.yml`** — a reusable that runs `tests/**/*.sh` and `tests/**/*.py`, reports each failing file as an annotation, and warns when a repo ships scripts but matched no tests. `require_tests: true` promotes that warning to a failure; the default keeps adoption non-breaking. `tests-caller.yml` pins the local copy, matching the `eval-validate-caller` convention.

**`validate-skill.sh`** — two new warnings:

- a script under `skills/*/scripts/` that no file under `tests/` references;
- an `llm_reviews` checkpoint whose prompt opens a line with a runnable command. This is `add-checkpoints`' own rule ("LLM reviews: for subjective requirements *that can't be mechanically verified*") made checkable — `git-workflow`'s GW-21 shipped an entire shell pipeline inside an `llm_review`, where nothing could execute or regress it. An entry that deliberately keeps both halves declares `# mechanical-counterpart: <ID>` and is exempt; prose that merely mentions a command inline is not a hit, since the command must start the line.

**`tests/mechanical-coverage.sh`** covers both checks, including the marker in either position and the inline-mention case. Three mutations were injected — drop the exemption, match commands mid-line, downgrade the warning to OK — and each fails a specific assertion.

## Dogfooding

The repo now satisfies its own rule. `tests/validate-evals.sh` and `tests/migrate-licensing.sh` cover the two scripts that had none. `validate-evals.sh` runs in every consumer's CI through the eval-validate reusable and had no test here, so a format regression would have surfaced first in a foreign repo. `migrate-licensing.sh` deletes a file and rewrites a README; its test pins the behaviour that matters legally — a foreign copyright holder is preserved and only the year extended, never replaced.

## Follow-up

Distributing `tests.yml` to consumers means adding a caller to `netresearch/.github/templates/skill/`, which is a separate PR against a different repo.


CybotTM:
Correction to this PR's rationale, found while adopting the reusable elsewhere.

The description says "this repo offers 20 reusable workflows and not one of them runs `tests/`". That is wrong. `ci-python.yml` runs a caller-supplied `test-command`, and at least one repo uses it exactly for this — retro-skill passes `python -m unittest discover -s tests -v` and has been gating its four suites all along. My fleet check grepped for `tests/` with a slash and missed `-s tests`.

The accurate statement: **no reusable ran shell tests**, and the Python path required each caller to hand-write its own command, so it was opt-in and easy to omit. The 276-scripts-without-tests measurement stands — those repos have no test files at all, whatever the pipeline could run.

What this changes for the code: nothing. `skill-repo-skill`'s own six suites are shell scripts that `unittest discover` cannot reach, and they were unexecuted until this PR. What it changes for the reasoning: the gap was narrower than I stated, and a repo already calling `ci-python.yml` with a test command does not need this reusable — netresearch/retro-skill#60 was opened on that mistaken premise and is closed.

PR #225, opened 2026-08-15
feat: compile assertion patterns and let evals self-check with samples

CybotTM:
Structural validation accepted any non-empty string as an assertion pattern, so two defects passed it unseen: a pattern that is not parseable at all, and a pattern that is parseable but inverted — one demanding a word the intended answer never contains.

The second is not hypothetical; it shipped in git-workflow-skill. An assertion there required the literal `remove` or `delete`, while the reference answer says "never-removable" and "not a removal". A correct answer failed the eval and a wrong one could pass it, and CI could not see any of it, because assertions were never executed against anything.

**Matching uses the grader's own instrument.** `run-ab-evals.sh` grades with `grep -qiE` after stripping a leading `(?i)`. The first version of this branch validated with Python's `re`, which disagrees in ways that decide real cases: `re` is case-sensitive, so an eval whose passing sample differs only in case was rejected although the grader accepts it, and in ERE `[^\n]` excludes the letter `n`, so this branch's own showcase pattern matched in one engine and not the other. Both checks now call `grep` with the grader's flags. For the same reason every assertion carrying `value` or `pattern` is covered, whatever its type — the grader has no type filter either, so a broken pattern under `tool_use` or `content_regex` is graded and was being skipped. A `must_not` assertion is checked inverted.

An eval may carry an optional `samples` block: one `passing` answer that every assertion has to match, and `failing` answers that at least one assertion has to reject. Deterministic, no model call — it decides mechanically the part of the documented quality rule ("every eval needs a delta-discriminating assertion") that is decidable without an A/B run. A block that is malformed, misspelled, or attached to an eval with no pattern used to be a silent no-op; each is now an error naming the wrong key.

**Blast radius, measured against `main` file by file across all 37 `evals.json` in the local fleet: zero changed verdicts.** Three assertions in two consumer repos are not parseable as ERE — `((`, `((.:var`, `public function __construct(` — and the grader can therefore never match them. Those are real defects, but they predate this change and sit under a `*_contains` type where the value is meant as a literal, so they warn and name the cause rather than turning a foreign repo's CI red.

One note on the tests, because the first version of them was green for the wrong reason: the validator enforces a minimum of 10 evals and 2 assertions per eval, so a fixture holding only the eval under test fails on those minimums instead of the defect being tested, and every "must fail" assertion passes vacuously. `suite()` pads each fixture, a separate case asserts the padding validates on its own, and the assertion that a broken pattern is *named* now matches the FAIL line rather than the eval name, which appears in the PASS output too.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_012NiLDH3iWw8CVdAnimJbF8)_

### DIFF
--- a/skills/skill-repo/references/materialization-contract.md
+++ b/skills/skill-repo/references/materialization-contract.md
@@ -179,13 +179,24 @@ decides it mechanically, with no model call:
 }
 ```
 
-Every `content` assertion must match `passing`, and each entry in `failing` must be
-rejected by at least one assertion. The block is optional and evals without it validate
-exactly as before, so adopting it never breaks a repo on day one. Write the `failing`
-samples from the answer a model actually gives without the skill — a straw man that
-shares no vocabulary with a correct answer passes the check while proving nothing.
-`content` assertion patterns are now compiled during validation as well, whether or not
-samples are present.
+Every assertion must match `passing`, and each entry in `failing` must be rejected by at
+least one assertion. Write the `failing` samples from the answer a model actually gives
+without the skill — a straw man that shares no vocabulary with a correct answer passes
+the check while proving nothing.
+
+**Matching uses the grader's instrument, not a similar one.** `run-ab-evals.sh` grades
+with `grep -qiE` after stripping a leading `(?i)`: case-insensitive POSIX ERE, with no
+type filter — it takes `value or pattern` from *every* assertion. The validator calls the
+same binary with the same flags, because Python's `re` disagrees with ERE in ways that
+decide real cases (`[^\n]` excludes the letter `n` in ERE, and `re` is case-sensitive),
+and a self-check measuring with a different engine certifies evals the grader would fail.
+For the same reason a pattern under `tool_use` or `content_regex` is validated too. An
+assertion typed `must_not` is checked inverted: it must NOT match `passing`.
+
+Two things change for evals **without** a samples block: patterns are now checked to be
+parseable by `grep -E`, and an unparseable one fails — except under a `*_contains` type,
+where the value is usually a literal and the mismatch is the grader's, so it only warns.
+Nothing else about validation changed, and no evals.json in the fleet changes verdict.
 
 ## Rule 8: Per-private-repo confirmation
 
--- a/skills/skill-repo/references/materialization-contract.md
+++ b/skills/skill-repo/references/materialization-contract.md
@@ -160,6 +160,33 @@ This verification is a **LOCAL/manual step, not CI** — `run-ab-evals.sh` calls
 `claude -p` per eval. It is referenced only by `.github/workflows/ab-evals-schedule.yml`,
 which is disabled by team decision, so it is not part of any active CI gate.
 
+### `samples`: the part of that quality rule CI can decide
+
+An assertion is only a non-empty string as far as structural validation used to be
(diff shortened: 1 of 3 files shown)


CybotTM (commented):
Self-review, because Copilot has no quota left until the monthly reset.

The decisive question for a validator change is whether it breaks anything in foreign repositories — this script runs in every consumer through the `eval-validate` reusable. Measured rather than assumed: the new validator was run against all 37 `evals.json` files in the local fleet and compared exit code by exit code against the previous validator. Not a single difference; the 7 files that fail, fail on both versions and for the same reason (eval count below the minimum).

The tests were written before the implementation, and the first version of them was green for the wrong reason: the validator enforces a minimum of 10 evals and 2 assertions per eval, so my minimal fixtures failed on those minimums instead of the defect under test, and every "must fail" assertion passed vacuously. That is corrected in the diff — `suite()` pads each fixture, and a separate case asserts the padding validates on its own. After that, exactly the three new cases failed, and they pass after the implementation.

One detail that went wrong while writing and is fixed in the diff: `suite()` first passed the subject through stdin, which cannot work — python's own heredoc owns stdin there, so `sys.stdin.read()` hands back this script. It travels through the environment now.

`tests/validate-evals.sh` is executed by `tests-caller.yml` through the `tests/**/*.sh` glob, so the new cases are an actual gate rather than a file that merely exists.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_012NiLDH3iWw8CVdAnimJbF8)_

CybotTM (commented):
Adversarial review across five dimensions, one refuter per finding defaulting to REFUTED: 29 findings over the three PRs of this round, 27 confirmed, 2 refuted, none left unverified. Copilot remains unavailable (monthly quota), hence this documented self-review.

The finding that mattered most invalidated the premise of the first version: the self-check matched with Python's `re` while the grader, `run-ab-evals.sh`, uses `grep -qiE`. A check on a different instrument than the thing it certifies is worse than no check, and the reviewer demonstrated both directions — an eval whose passing sample differed only in case was rejected although the grader accepts it, and this branch's own showcase pattern `\bswitch\b[^\n]{0,30}\b(main|master)` matched "switch onto main" in Python and not in ERE, because ERE's `[^\n]` excludes the letter n. Both checks now call grep with the grader's own flags.

Related and equally real: the type filter. The grader takes `value or pattern` from every assertion with no filter at all, so an unparseable pattern under `tool_use` or `content_regex` is graded and was being skipped — this repository's own evals.json carries exactly such an assertion. All of them are covered now, with `must_not` checked inverted.

Three further defects, all fixed: a samples block that was malformed, misspelled or attached to an eval with no pattern was a silent no-op; a non-list `failing` value raised a traceback out of the validator; and one of my own new tests was green for the wrong reason — it asserted the eval name appears, which it does in the ordinary PASS output too.

Blast radius re-measured after every change, file by file against `main` across all 37 evals.json in the local fleet: zero changed verdicts. Three assertions in two consumer repos are genuinely unparseable as ERE and can never match, but they predate this change and their type says the value is a literal, so they warn and name the cause instead of turning a foreign repository's CI red on its next pull request.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_012NiLDH3iWw8CVdAnimJbF8)_

PR #227, opened 2026-08-17
fix(fleet-release): let a prepared FIRST-RELEASE tag its committed version

CybotTM:
The monotonicity check demanded `version > .last` for every class except `TAG-ONLY`. For `FIRST-RELEASE` that is the wrong comparison: the repo has no release **and** no tag, so `.last` is not a shipped version at all — it is the version already committed in `.claude-plugin/plugin.json`. There is nothing to regress from, and tagging exactly that version is the normal case for a repo somebody prepared and never released.

Refusing equality made that state unrepresentable. The 2026-08-17 sweep hit it on both `ecom-*` repos and the only ways forward were to retype the row `TAG-ONLY` by hand or to invent a version nobody wrote — while the bump phase had handled the case correctly all along (`already at X ... finish will tag`).

`FIRST-RELEASE` now accepts `version >= committed` and refuses only a version **below** it, where parity would fail at tag time. `BUMP` still requires strictly greater — there a released version exists to regress from — and `TAG-ONLY` still requires exact equality.

### Verification

Four new cases around the new branch, plus the probe that matters: with the old logic restored, only `may tag the already-committed version` fails and the other three still pass. So one assertion genuinely discriminates for this fix, and `below the committed version refused` guards the tempting over-broad version of it (accepting anything for the class). 95 tests pass, shellcheck clean, and the behaviour matrix was re-checked through the **vendored** copy in `gitlab-skill` as well, where both existing guards stay intact.

The vendored engine in `netresearch-gitlab` is re-synced in a companion MR, so the two hosts do not diverge.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01UDJDghtNvgEJgQX8StHHjf)_


CybotTM:
Copilot review is unavailable — the account-wide monthly quota is exhausted, so re-requesting cannot change it this month. Reviewed it myself instead, and recorded the one thing worth recording: of the four new tests, only `may tag the already-committed version` actually fails against the old logic. The other three pass either way, so they are guard rails rather than regression tests for this change — `below the committed version refused` is the one that would catch the tempting over-broad fix (accepting any version for the class), which is why it stays.

Checked beyond the suite: 22 CI checks green, shellcheck clean, and the behaviour matrix re-run through the vendored copy in `netresearch-gitlab` rather than inferred from this one — `FIRST-RELEASE` equal/above accepted, below refused, `BUMP` equality still refused, `TAG-ONLY` inequality still refused.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01UDJDghtNvgEJgQX8StHHjf)_

Issue #148, opened 2026-07-10
A6: Eval backfill + quality rule — every eval must assert the model delta


CybotTM:
Part of the skill value-gate program — epic: https://github.com/netresearch/skill-repo-skill/issues/142. **Priority** P2 · **Effort** L · **Verification** NEEDS_REVISION (corrections folded in below) · Wave 4 · Coordinates per-repo work.

## Quality rule (write into the eval-integration guidance here)

Every eval needs at least one assertion a no-skill baseline answer FAILS; verify with `run-ab-evals.sh` (the `without_skill` pass rate on that assertion must be <1.0). Current keyword assertions ("alpine", "USER", "Plugin") reward boilerplate and never test the retro-born gotchas that are the skills' actual value.

## Backfill checklist (public catalog; each item = small PR in the named repo)

- [x] docker-development-skill: eval for the healthcheck-disable/`--wait` trap (**correction**: the pgrep `[m]essenger` guard eval already exists as `worker-healthcheck-pgrep-self-match` — only the healthcheck eval is missing); new eval for the docker-via-wsl skill (repo's evals.json holds 24 docker-development evals, zero for docker-via-wsl) asserting the `wsl.exe -e bash -lc` re-issue.
- [x] typo3-site-conformance-skill: first evals.json (none exists).
- [x] jujutsu-workflow-skill: first evals.json (none exists).
- [x] typo3-upgrade-effort-model-skill: first evals.json — deterministic formula outputs make good assertions.
- [x] markdown-to-pdf-skill: smoke eval (PDF exists, size >0; none exists).
- [x] typo3-a11y-skill: eval asserting generated nav does NOT contain `role="menu"` (anti-pattern documented at SKILL.md:24).
- [x] typo3-vite-skill: eval asserting exactly one `server` block + `allowedHosts` (gotcha at references/vite-configuration.md:98-112).
- [x] typo3-ckeditor5-skill — **correction**: the proposed upcast/downcast eval already exists (`upcast-downcast-converters`); the "Plugin" keyword lives in the separate `create-ckeditor5-plugin` eval. Review that one instead of adding a duplicate.

Two further backfill targets lived in the internal catalog, tracked as `coding-ai/marketplace#3`; both are done and merged (`coding-ai/gitlab-skill!79`, `coding-ai/netresearch-jira-skill!97`).

[...]

PR #229, opened 2026-08-17
feat(evals): bind every A/B delta to the actor that produced it

CybotTM:
Part of [#148](https://github.com/netresearch/skill-repo-skill/issues/148) — the rule and the instrument, not yet the per-repo backfill.

A delta is not a property of a skill:

```text
delta = f(skill version, model, agent harness, task, tool environment)
```

So `docker-development has +23%` is not a claim anyone can check, and a stronger future actor can erase the same delta while a discovery-weak harness can suppress it with the skill unchanged. Two things follow.

**1. Every run records and prints the actor tuple.** `provenance` gains `harness` (from `claude --version`), `skill_version` (nearest `plugin.json`/`composer.json` above the SKILL.md) and `eval_set` (path, content id, count) — so two runs quoting one number can show they graded the same questions. The run ends with:

```text
Measured: docker-development@2.4.0 · claude-code 2.0.31 · model sonnet (judge haiku) · eval-set evals.json@09cc52ef (24 evals)
Discriminating checks (baseline fails, skill passes): 14
Quote the delta with that tuple, not on its own.
```

`merge-ab-results.py` carries the fields into the dashboard entry, and the dashboard renders them: the actor beside the aggregate in the header, the whole tuple on every delta. The 25 existing entries have no provenance at all, so they read **"harness not recorded"** — accurate, and the point: those numbers were published without recording what produced them.

**2. The runner computes the A6 rule instead of only asking for it.** `materialization-contract.md` already requires every eval to have an assertion the no-skill baseline fails; nothing measured it. A *discriminating check* is now defined as one the baseline fails and the skill run passes, reported per eval as `discriminating=<n>/<total>`, with every zero-scoring eval listed by name and both facts in `ab-results.json` (`discriminating_checks`, `evals_without_delta`). `--require-delta` turns that list into exit 1 — for repos whose evals are already clean, which is what the backfill half of #148 has to produce first. Not wired into any CI job in this PR, deliberately.

Zero discriminating checks does not make an eval wrong — a regression guard both arms pass is doing its job — but it cannot be evidence *for the skill*, which is the claim being made whenever skill content is defended.

Docs follow ownership rather than duplicating: the rule stays canonical in `materialization-contract.md` (now with the tooling), and `skill-quality.md` owns the reporting side plus why the two eval levels answer different questions — the per-skill A/B assumes the skill is loaded and structurally cannot see a discovery failure, which is what [`netresearch/agent-system-evals`](https://github.com/netresearch/agent-system-evals) measures.

**Verified locally:**

- `tests/run-ab-evals.sh` — new, 16 assertions against a `claude` stub (no model calls): the tuple, per-eval counts in both the regex and the LLM-graded path, the named evidence-free eval, and both exit codes of `--require-delta`. It runs in CI via `tests-caller.yml` → `tests.yml` (`tests/**/*.sh`).
- Defect-injected: inverting the discrimination condition fails six assertions including both `--require-delta` directions, then restored byte-identical and green again.
- All 8 pre-existing `tests/*.sh` suites pass, `pre-commit run --files <changed>` green (`shellcheck`, `markdownlint`, `ruff`), `validate-skill.sh` and `validate-evals.sh` clean, `audit-skills.sh` unchanged verdicts.
- Dashboard driven in a real browser against the committed `results.json`: 25 rows, no page errors, header reads `model opus · harness not recorded`, delta tooltips read `netresearch-branding@2.6.0 · harness not recorded · model opus · 21 evals`.

### DIFF
--- a/docs/dashboard/index.html
+++ b/docs/dashboard/index.html
@@ -394,7 +394,13 @@ async function loadData() {
     // Result blocks are keyed by the model id recorded in provenance
     // (legacy entries: "opus"). Alias the single model key to `opus`
     // so the render code below has one canonical accessor.
-    DATA.skills.forEach(s => { if (s.results && !s.results.opus) s.results.opus = Object.values(s.results)[0]; });
+    DATA.skills.forEach(s => {
+      // Keep the model the entry was actually measured under before the alias
+      // below adds a second key; the provenance block is authoritative when
+      // present, and for older entries the result key is all there is.
+      s.modelKey = (s.provenance && s.provenance.model) || (s.results && Object.keys(s.results)[0]) || null;
+      if (s.results && !s.results.opus) s.results.opus = Object.values(s.results)[0];
+    });
   } catch (e) {
     document.body.innerHTML = '<p style="padding:2rem;color:red;">Failed to load data/results.json</p>';
     return;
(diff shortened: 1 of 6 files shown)


CybotTM:
**Self-review** — Copilot is out of its monthly review quota, so no bot review is available. Reviewed the diff myself; recording what I checked, since I also wrote it.

- **The gate this adds runs, and it ran here.** `tests / Skill Tests` executed `tests/run-ab-evals.sh` on this head: [16 assertions, all green](https://github.com/netresearch/skill-repo-skill/actions/runs/32060848303/job/95481558975). No model is called — the fixture puts a `claude` stub on PATH — so the job stays cheap and deterministic.
- **Nothing existing changes value.** The regex and LLM pass rates are computed exactly as before; `disc` is counted alongside them and never feeds a total. Published numbers do not move.
- **`--require-delta` is opt-in and wired nowhere**, deliberately: the per-repo eval backfill in #148 has to land first, otherwise the flag turns repos red for evals nobody has revisited yet.
- **Shell details checked:** the `claude --version` substitution carries its own `|| true`, the fall-through `[[ … ]] && …` line was exercised in both directions during testing, and the empty-array expansion under `set -u` is covered by the clean-run case in the test.
- **`merge-ab-results.py` re-run by hand** against a synthetic `ab-results.json`: the entry carries `provenance.harness`, `skill_version` and `eval_set`, and the run prints `demo@4.5.6: … model=sonnet harness=claude-code 9.9.9` plus the evidence-free eval by name.
- **Dashboard driven in a real browser** against the committed `results.json`: 25 rows, no page errors, header `model opus · harness not recorded`, tooltips `netresearch-branding@2.6.0 · harness not recorded · model opus · 21 evals`. Existing entries read "not recorded" because they were published without recording the actor — accurate, and the reason for the change.
- **Docs follow ownership:** the eval rule stays canonical in `materialization-contract.md` (now with the tooling that computes it), `skill-quality.md` owns the reporting side and links across rather than restating.

Checks: 22 pass, 0 fail, 2 skip (of 24). Threads: 0. `mergeState=CLEAN`.

Merging on that basis.

Issue #232, opened 2026-08-19
Arm --require-delta per repo: measure each suite first, then make the eval rule a gate


CybotTM:
[#229](https://github.com/netresearch/skill-repo-skill/pull/229) added the instrument: `run-ab-evals.sh` computes a **discriminating check** per eval (baseline fails it, skill run passes it), prints `discriminating=<n>/<total>`, lists every eval that scores zero, records both in `ab-results.json`, and exposes `--require-delta` as an exit code. [#148](https://github.com/netresearch/skill-repo-skill/issues/148) completed the public backfill. What is still a convention rather than a gate is the last step: nothing runs the flag.

**Why it cannot simply be switched on.** No suite in the catalog has ever been measured against it. The rule was checked by hand until now, so an unknown number of evals pass in both arms — and a repo whose `main` turns red on a check nobody has seen the output of gets the check disabled rather than the evals fixed. The order has to be: measure, fix or accept per eval, then arm.

**Per repo:**

1. Run `scripts/run-ab-evals.sh --evals=<repo>/evals/evals.json --skill=<repo>/skills/<name>/SKILL.md` and read the `evals_without_delta` list.
2. For each eval on it, decide: rework the assertion (it measures boilerplate), or record why it stays (a regression guard both arms pass is legitimate — it is just not evidence *for* the skill).
3. Only then add `--require-delta` to whatever invokes the runner for that repo.

**Two things to settle before the first repo.** The runner calls `claude -p` per eval, so this costs real tokens and cannot live in ordinary PR CI — `ab-evals-schedule.yml` is the existing home and is disabled by team decision, which means the gate needs either a decision to re-enable it or a local-only convention with a documented cadence. And [#230](https://github.com/netresearch/skill-repo-skill/issues/230) has to land first: while `must_not` is graded as a positive match, any eval using that type reports a wrong discriminating count, so arming the flag would fail the wrong evals.

Filed because this was living in a comment on #148, which is about the backfill — its public checklist is done, and this step is a different piece of work.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01CT41JfSGYEJJaBUZxg7xzu)_

CybotTM:
Closing. The issue asked to measure each suite and then arm `--require-delta`. Measuring answered the second half differently than expected, so here is what was done and what the evidence says.

## The instrument was not measuring the skill

[...]

PR #233, opened 2026-08-19
fix(evals): grade must_not assertions in the direction they state

### LINKED ISSUE #230: run-ab-evals.sh grades must_not assertions as positive matches
`run-ab-evals.sh` takes `value or pattern` from **every** assertion with no type filter and counts a `grep -qiE` hit as a pass. An assertion typed `must_not` therefore scores backwards: the arm that contains the forbidden phrase is credited with a pass, so a baseline answer giving exactly the wrong advice raises the `without` rate and shrinks the delta.

`validate-evals.sh` already handles the type correctly — it checks a `must_not` pattern inverted against `samples.passing` — so the two instruments disagree about what the same eval means, and the one that decides published numbers is the wrong one.

Consequence today: an eval cannot express a negative expectation to the A/B runner at all. In [netresearch/docker-development-skill#62](https://github.com/netresearch/docker-development-skill/pull/62) the negative case had to move into `samples.failing` and the `must_not` assertion be dropped, because leaving it in would have corrupted the delta rather than tightened it.

Fix is small and local to the grading loop: read the assertion type alongside the pattern (the loop currently emits only `value or pattern`), and invert the hit for `must_not`. It moves existing numbers for any eval that uses the type, so it wants a re-run of the affected repos rather than a silent flip — worth checking how many evals across the catalog carry `must_not` before deciding whether that is one repo or twenty.

Related: [#229](https://github.com/netresearch/skill-repo-skill/pull/229) added per-eval discriminating-check reporting, which reads the same grading result and inherits the same inversion.

CybotTM:
Closes #230.

The grading loop emitted only `value or pattern`, so the assertion type never reached the comparison and everything was scored as "pattern present = pass". For `must_not` that is backwards: the arm giving the forbidden answer got the point, the baseline rate rose, the delta shrank, and the discriminating count added in #229 inherited the same inversion. A negative expectation could not be stated as an assertion at all — in [docker-development-skill#62](https://github.com/netresearch/docker-development-skill/pull/62) it had to move into `samples.failing`.

**The fix** reads kind and pattern as a two-field line — kind first and never empty, so `IFS` cannot collapse it into the pattern field — and inverts the verdict for `must_not`. Passing means the pattern is **absent**; a discriminating check means the baseline produced the forbidden answer and the skill run did not.

**Blast radius, measured rather than assumed.** 58 `must_not` assertions in two repos:

| repo | assertions | published delta |
|---|---|---|
| german-technical-writing-skill | 35 | none |
| github-release-skill | 23 | none |

Neither appears in `docs/dashboard/data/results.json`, so no published number was computed in the wrong direction and nothing has to be retracted. Both want a run before their first publish — which is [#232](https://github.com/netresearch/skill-repo-skill/issues/232)'s job anyway, and #230 was its stated blocker.

**Docs corrected.** Three places stated "no type filter" as settled behaviour: the validator header, the validator's inline justification for checking every pattern, and `materialization-contract.md`. They now separate the two questions — every assertion is graded whatever its type, and the type decides the *direction* of the verdict.

**Verified:**

- `tests/run-ab-evals.sh` gains the mirror case, and I ran it against the **unfixed** grader first: three assertions fail, including "an eval whose only movement is a `must_not` violation is not counted as evidence" (it was credited with a discriminating check before).
- All 9 `tests/*.sh` suites pass; `pre-commit run --files <changed>` green; `shellcheck` clean.

### DIFF
--- a/scripts/run-ab-evals.sh
+++ b/scripts/run-ab-evals.sh
@@ -330,7 +330,10 @@ print(len(evals[$i].get('assertions', [])))
 ")
     if [[ "$assertion_count" -gt 0 ]]; then
         pass_w=0; pass_s=0; total=0; disc=0
-        while IFS= read -r pattern; do
+        # Each line is "<kind><TAB><pattern>", kind being must_not or positive.
+        # The kind is emitted first and is never empty, so IFS cannot collapse
+        # it into the pattern field.
+        while IFS=$'\t' read -r kind pattern; do
             [[ -z "$pattern" ]] && continue
             # Strip PCRE inline flags (e.g. (?i)) — grep -i already handles case
             pattern="${pattern#'(?i)'}"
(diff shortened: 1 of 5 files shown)


CybotTM:
**Self-review** — no bot reviewer available (Copilot quota is out for the month). Reviewed the diff myself; recording what I checked, since I also wrote it.

- **The test fails against the unfixed grader.** I ran it that way before committing: three assertions fail, including "an eval whose only movement is a `must_not` violation is not counted as evidence" — before the fix that case was credited with a discriminating check. All 19 assertions pass after, and `tests / Skill Tests` executed the file on this head.
- **The two-field line cannot collapse.** The kind is emitted first and is never empty, so `IFS=$'\t' read -r kind pattern` has nothing to shift; the pattern keeps whatever tabs or spaces it contains because it is the last field.
- **Existing numbers do not move.** Only assertions typed `must_not` change verdict, and the blast radius is measured rather than assumed: 58 of them in `german-technical-writing-skill` (35) and `github-release-skill` (23), neither of which has an entry in `docs/dashboard/data/results.json`. Nothing published was computed with the wrong direction.
- **The docs no longer state the old behaviour as settled.** Three places claimed "no type filter"; they now separate *whether* an assertion is graded (always) from *which direction* the verdict runs (the type decides).
- All 9 `tests/*.sh` suites pass, `shellcheck` clean, `pre-commit run --files <changed>` green.

Checks: 22 pass, 0 fail, 2 skip (of 24). Threads: 0. `mergeState=CLEAN`. Closes #230.

Merging on that basis. The follow-up it unblocks is #232, and running the suite against the fixed grader immediately turned up an instrument problem worth its own change — reported there rather than smuggled in here.

PR #234, opened 2026-08-19
fix(evals): isolate the A/B measurement, and report evidence-free evals

CybotTM:
Refs #232. The issue asks to measure each suite and then arm `--require-delta`. Running the first measurement turned up two instrument defects, so this PR is the prerequisite: a measurement worth gating on.

**What the first two runs actually produced.** A plausible-looking aggregate (`+23` points on skill-repo's own suite) with two evals at a *negative* delta. Reading the raw arms — `scripts/ab-results/<eval>_with.txt`, which nothing in the pipeline shows you — explained both:

1. **`--tools ""` drops the built-in tools and nothing else.** The operator's MCP servers stay attached, so answers on this workstation discussed which tools were available (Gmail, Drive) instead of the question — and unequally between the arms. Fixed with `--strict-mcp-config` and an empty `--mcp-config`.
2. **With the tools gone the model still expects them.** An action-shaped prompt (`Create a valid plugin.json …`) was answered with *"I'll check the existing repo structure first"* plus a tool call print mode cannot serve: **3 output tokens**, nothing to grade. The negative deltas were the with-skill arm spending its answer explaining the convention while the baseline happened to emit more literal matches. Both arms now carry a note that there are no tools, so they still differ only in the skill.

Measured on the eval that produced the 3-token answer:

```text
before: I'll check the existing repo structure first.  *[Tool: bash]*
after:  Since I have no file tools in this session, here's the content — … {"name": "security-audit", …}
```

**The sweep now reports what the runner computes.** Every skill's `evals_without_delta` list lands in the refresh PR body under its own heading (`scripts/no-delta-line.py`). Deliberately a *report*, not a gate: the sweep publishes numbers, and a gate in its loop would abort the merge step for that skill — one weak eval and a skill quietly stops being refreshed on the dashboard while looking merely unchanged.

**`skill-quality.md`** gains three things: where the gate belongs (local, pre-merge, per repo — it costs two model calls per eval, so it cannot be PR CI), why CI should prefer the model-free `samples` check instead, and the read-the-arms rule that found both defects above.

**Verified:**

- `tests/run-ab-evals.sh` asserts every completion call carries the isolation flags — verified by deleting one flag, which fails it — and the stub now identifies the arm by whether the appended system prompt contains the skill body, since both arms pass `--append-system-prompt` now.
- All 9 `tests/*.sh` suites pass; `pre-commit run --files <changed>` green (`actionlint`, `yamllint`, `shellcheck`, `ruff`, `markdownlint`).
- Re-measured end to end against the fixed runner; the with-arm answers now contain the artefact the assertions grade instead of tool talk.

---

**Third defect, found by the re-measurement itself** (second commit): the run hit the account's session limit on its last eval pair. The CLI writes that notice to stdout, the runner redirects it into the answer file, and the grader scored it — 0 of 4 assertions in both arms, and the eval landed in the evidence-free list as if its assertions were weak. Nothing in the output distinguished it from a genuine both-arms-pass result.

A pair whose arm carries no answer is now **NOT MEASURED**: excluded from the totals, listed separately as `evals_unmeasured`, and skipped by `--require-delta` — unknown is not refuted, and the difference decides whether someone rewrites a perfectly good eval. The flag also gained a floor, because a fully blocked run would otherwise report zero evidence-free evals and exit 0.

Detection covers session/usage limits, credit-balance and rate-limit errors, invalid keys, `overloaded_error`, and an empty file after a timeout. Verified by neutering the detector: four assertions fail, including the one that keeps a blocked eval out of the evidence-free list.

---

**Two more commits, both driven by what the measurement showed.**

**Eleven of this repo's own evals reworked.** The first sound measurement put 11 of 20 measured evals in the evidence-free list, and reading them explained it: each asserted words its own prompt already contained (`(no|not|never)` + `git tag` for "should composer.json carry a version field?"; `Installation`, `License`, `Netresearch` for the README question). They now assert what only this skill supplies — the plugin.json/tag parity `check-plugin-version.sh` enforces, the split-licence pair, the 500-word count being taken over the whole file including frontmatter, org reusables staying on `@main` while SHA-pinning applies to third-party actions, the exact level-2 README headings with their three-prompt minimum, `sed -i` differing between GNU and BSD because the matrix includes macOS.

Same actor tuple both times (`skill-repo@1.35.0 · claude-code 2.1.235 · sonnet`):

| | before | after |
|---|---|---|
| baseline | 40/65 | 44/80 |
| with skill | 58/65 | 60/80 |
| discriminating checks | 18 | 23 |
| of the 11 reworked, still evidence-free | — | 6 |

**And the result that decides how the flag may be used.** Two evals I did *not* touch flipped from discriminating to not between the two runs:

| eval | run 1 | run 2 |
|---|---|---|
| `migrate_single_license` | 2 | 0 |
| `release_workflow_setup` | 1 | 0 |

Each arm is a single completion, so per-eval discrimination is noisy: `--require-delta` would have failed those two on one day and passed them on the other, with nothing about them changed. It is therefore an **opt-in local signal** — read the list, open the two arms, decide — and not a hard gate. Turning it into a gate needs repeated sampling per arm at a multiple of the token cost, which is a budget decision rather than a flag, and is written down as such in `skill-quality.md`.

SonarCloud's six findings on this branch are fixed in their own commit (three `S7677` diagnostics to stderr, the `S8707` taint validated at the path boundary, `S1192`, `S7688`).

**The six that still do not discriminate**, named rather than counted, so the next person does not have to re-measure to find them: `composer_no_lock_file`, `skillmd_word_limit`, `reusable_workflow_caller`, `cross_platform_scripts`, `fix_validation_errors`, `readme_requirements`. Plus `retire_superseded_skill`, which was blocked by the session limit in the earlier run and measured evidence-free in this one. They are left as they are: each asserts a fact the skill genuinely owns, and a baseline that happens to produce the same words is a statement about the model, not a defect in the eval — which is exactly the judgement the flag is meant to prompt rather than pre-empt.

---

**Correction to a claim in the commit message of `b331f21`.** It says the exec bit is invisible locally because "WSL reports every file as executable". Measured here, that is wrong on this machine: `~/p` is native ext4 (`/dev/sdd`), a file written by an editor is `0o644`, `ls -l`, `stat` and `os.access(X_OK)` all agree it is not executable.

What *is* reproducible is the discrepancy itself, and only that: `uvx --no-build ruff@0.16.0 check --extend-exclude .skill-repo-tools .` — the exact command and version from `validate.yml` — passes locally on `scripts/no-delta-line.py` at mode `644`, while the same command on the runner reports `EXE001`. Ruff does inspect the file locally (it reports `I001` on a copy with an unsorted import), and `EXE001` is in `linter.rules.enabled` in the local settings dump. So the rule is enabled, the file is genuinely non-executable, and the finding still does not appear here.

I could not establish the mechanism, so I am not going to name one. The operative fact for anyone hitting this: **a local ruff run cannot clear `EXE001` on this setup — check the committed mode with `git ls-tree HEAD <file>` instead** (`100755` here).


CybotTM:
**Self-review** — no bot reviewer available (Copilot quota is out for the month). Reviewed the diff myself; recording what I checked, since I also wrote it.

**What changes behaviour, and what only reports.** The runner changes are all in the two completion calls and the analysis loop: MCP isolation, the shared no-tools note, and the non-answer detection. Nothing touches how a matched assertion is scored, so a suite with none of these conditions produces the same numbers as before. The sweep change adds a section to a PR body and cannot fail the run — `no-delta-line.py` is called with `|| true` and stays silent when the list is empty.

**The two arms still differ only in the skill.** The no-tools note is appended to both, and the test's stub now identifies the arm by whether the appended prompt contains the skill body rather than by the presence of `--append-system-prompt`, which both arms now pass.

**Every claim in the PR body is measured, not argued.** Three full runs of this repo's own suite, same actor tuple, with the raw arms read in each case. The before/after on the reworked evals (18 → 23 discriminating checks) and the two flipped evals that decide against arming the flag both come from those runs, not from reasoning about them.

**Three things went wrong on the way here, all mine.** `ruff EXE001` on the new script — the committed tree now has `100755`; the mechanism claim in that commit message is wrong and is corrected in the PR body, because the local run genuinely cannot reproduce the finding and I could not establish why. A race in the test stub's log, where three separate `printf` calls interleaved across parallel stubs and produced a line without its prefix; the stub writes one record now, run three times in a row to confirm. And the commit that claimed that stub fix carried only the mode bit, because the `git add` sat in a command a hook rejected and a rejected command runs nothing — landed separately as `3574046`, which is why the isolation assertion now passes in CI (`ok every completion call carries the MCP isolation flags`) rather than reporting 8 against 9.

**Verified:** all 9 `tests/*.sh` suites pass, `uvx ruff check scripts/ tests/` clean, `validate-evals.sh` 45 passed / 0 failed / 0 warnings, `pre-commit run --files <changed>` green, and the defect-injection checks on both new behaviours (remove an isolation flag → the assertion fails; neuter the non-answer detector → four assertions fail).

Refs #232 — which this closes as "instrument sound, reporting continuous, gate deliberately not armed", with the measurement that forced that conclusion in the body above.

Merging on that basis.

PR #237, opened 2026-08-19
feat(evals): decide every assertion by majority over repeated samples

### LINKED ISSUE #236: Decide: repeated sampling per arm, or the eval rule stays reported rather than gated
[#232](https://github.com/netresearch/skill-repo-skill/issues/232) closed with the eval rule *reported* rather than *gated*, because the measurement cannot support a gate at one sample per arm. This is the decision that would change that, and it needs a number from a person rather than an implementation from me.

**The evidence.** Two runs of this repo's own suite, same eval text, same actor tuple (`skill-repo@1.35.0 · claude-code 2.1.235 · sonnet`), nothing changed between them:

| eval | run 1 | run 2 |
|---|---|---|
| `migrate_single_license` | 2 discriminating checks | 0 |
| `release_workflow_setup` | 1 | 0 |

`run-ab-evals.sh --require-delta` would have failed those two on one day and passed them on the other. A gate that flips on unchanged input fails evals for the wrong reason — the same failure mode [#148](https://github.com/netresearch/skill-repo-skill/issues/148) exists to prevent, one level up.

**What a gate would require.** Repeated sampling per arm, with the verdict taken over the samples (e.g. "discriminating in ≥2 of 3 runs"). The cost scales linearly: the current sweep is 2 completions per eval, so 3 samples per arm makes it 6. For this repo's 21 evals that is 126 completions instead of 42; across the ~25-skill catalog on the weekly rotation it multiplies the sweep's whole budget by three.

**The decision, stated as a question:** is a hard eval-quality gate worth 3× the sweep's token budget, or is the reported list enough? Both are defensible. The reported list already puts every evidence-free eval in front of a human on each rotation, and the mechanical half — the `samples` self-check, which needs no model at all — is already enforced in CI for every new eval.

**If yes**, the work is bounded: an `--samples N` flag on the runner, majority verdict per assertion, provenance recording N, and the sweep's PR body reporting the sample count alongside the list. **If no**, close this and the current arrangement stands.

Filed because it was living in a comment on a closed issue.

CybotTM:
Closes #236, answered yes: pay the multiplier, get a verdict that holds still.

## What changes

`--samples=N` runs N completions per arm and decides **each assertion by majority over them**. `must_not` inverts each sample before the vote, so validator and grader still agree on direction. Samples whose answer never arrived — session limit, API error, empty body — are dropped rather than counted as failures, and an arm left with none still makes the eval *unmeasured*. The LLM-graded path votes the same way, per expectation number.

`--require-delta` refuses to run below `--samples=2`. Gating on a single completion is the configuration #236 rejected; leaving it reachable would have kept that failure mode one flag away.

The sweep passes **3** by default and states it in the PR body it opens, so a reader knows how firm the per-eval verdicts are.

## The cost, stated plainly

`2 × N` completions per eval. For this repo's 21 evals that is **126 instead of 42**; across the weekly rotation it triples the sweep's bill. The number is in the workflow input description rather than in a comment, because it is the entire content of the decision this PR implements.

`--samples=1` remains available for a cheap indicative run — the aggregate still says something — but `skill-quality.md` now says not to quote a per-eval verdict from it.

## How the property is tested, not just the plumbing

The test stub makes exactly **one** with-arm sample miss an assertion, and the index is drawn through an atomic `mkdir` ticket, so it is deterministic without depending on how the parallel calls happen to be scheduled. Then:

| | verdict |
|---|---|
| `--samples=1` | the eval lands in `evals_without_delta` — one flaky answer decided it |
| `--samples=3` | the majority outvotes the flaky sample and it counts as evidence |

That is the flip from #236 reproduced in a fixture, and it is what the two new assertions pin. 36 assertions total, stable across three consecutive runs of the suite.

## Detail

Sample 1 keeps the historical filename; further samples get an `_sN` suffix, so existing eyeballs and tooling still find the first answer. `samples_per_arm` is recorded in `provenance` and printed in the `Measured:` line. Two helpers that only ever read a single grader file are removed, superseded by the majority function.

**Verified so far:** all 9 `tests/*.sh` suites pass, `shellcheck` and `actionlint` clean, `pre-commit run --files <changed>` green. A real 3-sample run against this repo's own suite is in flight; I will post its numbers here, in particular whether `migrate_single_license` and `release_workflow_setup` — the two evals that flipped between single-sample runs — now hold still.

---

## First sampled run against this repo's own suite

`skill-repo@1.35.0 · claude-code 2.1.235 · sonnet · 21 evals · 3 samples/arm` — 126 completions, 0 unmeasured:

| | single sample (previous run) | 3 samples |
|---|---|---|
| baseline | 44/80 | 43/80 |
| with skill | 60/80 | 65/80 |
| discriminating checks | 23 | 24 |
| evidence-free evals | 9 | 8 |

The two evals that motivated the issue:

| eval | run A (n=1) | run B (n=1) | sampled (n=3) |
|---|---|---|---|
| `migrate_single_license` | 2 | 0 | **2** |
| `release_workflow_setup` | 1 | 0 | **1** |

Both report the value they had in the run where they discriminated, which is what a majority should do with one outlying sample.

**That is one sampled run, so it does not yet establish stability** — a single measurement claiming "this is now stable" would repeat the error this PR exists to fix. A second sampled run is in flight; I will post the comparison, and if the per-eval verdicts differ between two sampled runs, that is a finding against `--samples=3` being enough and belongs here rather than in a later surprise.

### DIFF
--- a/.github/workflows/ab-evals-caller.yml
+++ b/.github/workflows/ab-evals-caller.yml
@@ -38,6 +38,13 @@ on:
         description: "Run the full catalog instead of the rotation subset (several hours)"
         type: boolean
         default: false
+      samples:
+        description: >-
+          Completions per arm (default 3). A verdict from one completion flips
+          between runs on unchanged evals, so 3 is the number the per-eval
+          results are quoted from; the run costs 2xN completions per eval.
+        type: number
+        default: 3
 
 permissions: {}
 
@@ -56,5 +63,6 @@ jobs:
     with:
       skills: ${{ inputs.skills || '' }}
       full: ${{ inputs.full || false }}
+      samples: ${{ inputs.samples || 3 }}
     secrets:
       ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
--- a/.github/workflows/ab-evals-schedule.yml
+++ b/.github/workflows/ab-evals-schedule.yml
@@ -32,6 +32,16 @@ on:
         required: false
         type: string
         default: ""
+      samples:
+        description: >-
+          Completions per arm. 3 is the default because a verdict from a single
+          completion flips between runs on unchanged evals (issue #236); the run
+          costs 2xN completions per eval, so 3 triples the token bill against
+          the old behaviour. Lower it to 1 only for a cheap indicative run whose
+          per-eval verdicts must not be quoted.
+        required: false
+        type: number
+        default: 3
       full:
         description: "Run the full catalog instead of the rotation subset (several hours)"
         required: false
@@ -151,6 +161,7 @@ jobs:
       - name: Run A/B evals per skill (fail-soft)
         env:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          SAMPLES_PER_ARM: ${{ inputs.samples }}
         run: |
           mkdir -p /tmp/ab-out
           : > /tmp/ab-out/pr-lines.md
@@ -181,7 +192,8 @@ jobs:
               failed+=("$name"); echo "::endgroup::"; continue
             fi
             rm -rf scripts/ab-results
-            if ! bash scripts/run-ab-evals.sh --evals="$evals" --skill="$skill" < /dev/null; then
+            if ! bash scripts/run-ab-evals.sh --evals="$evals" --skill="$skill" \
+                 --samples="${SAMPLES_PER_ARM}" < /dev/null; then
               echo "::warning::${name}: eval run failed, skipping"
               failed+=("$name"); echo "::endgroup::"; continue
             fi
@@ -240,6 +252,7 @@ jobs:
       - name: Open PR with refreshed dashboard data
         env:
           GH_TOKEN: ${{ github.token }}
+          SAMPLES_PER_ARM: ${{ inputs.samples }}
         run: |
           if git diff --quiet -- docs/dashboard/data/results.json; then
             echo "No changes to dashboard data; skipping PR."
@@ -262,6 +275,8 @@ jobs:
             echo ""
             if [[ -s /tmp/ab-out/no-delta.md ]]; then
               echo "## Evals with no discriminating check"
+              echo ""
+              echo "Verdicts are a majority over ${SAMPLES_PER_ARM} completions per arm."
               echo ""
               echo "The baseline answered these as well as the skill run, so they are not"
               echo "evidence for the skill. That does not make them wrong — a regression"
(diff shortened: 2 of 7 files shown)


CybotTM:
**Self-review** — no bot reviewer available (Copilot quota out for the month). Reviewed the diff myself; recording what I checked, since I also wrote it.

**The default does not change.** `--samples` defaults to 1, so an existing invocation costs exactly what it cost before and produces the same numbers — majority over one sample is that sample. Only the sweep opts into 3, and its input description carries the cost rather than hiding it.

**The gate cannot regress to a coin flip.** `--require-delta` exits 2 below `--samples=2`, so the rejected configuration is unreachable rather than merely discouraged. Two older tests that used the flag without sampling now pass `--samples=2`; that is a real behaviour change and it is the point of the issue.

**Non-answers still mean unknown.** Dropping happens per sample; an arm with no usable sample leaves the eval unmeasured, and a partly-dropped arm votes over what is left. A limit that hits one sample of three no longer discards the eval.

**The test proves the property, not the plumbing.** The stub makes exactly one with-arm sample miss an assertion, via an atomic `mkdir` ticket, so the fixture does not depend on scheduling. At `--samples=1` that eval is evidence-free; at `--samples=3` it is not. Run three times consecutively, stable.

**Dashboard commit:** the tuple beside every delta now names the sample count, and pre-sampling entries read "sample count not recorded" rather than being taken as firm. Driven in a real browser: 25 rows, no page errors.

**Measured, not argued:** a full 3-sample run against this repo's own suite (126 completions, 0 unmeasured) is in the PR body, including both evals that flipped between single-sample runs. A second sampled run is still going — its comparison is what would show whether 3 is enough, and it lands here either way. If it disagrees, the follow-up is a higher N, not a revert of this.

Checks: 22 pass, 0 fail, 2 skip (of 24). Threads: 0. `mergeState=CLEAN`. Closes #236.

Merging on that basis, per instruction.

CybotTM:
The second sampled run finished, and the comparison I promised goes **against** what this PR assumed: four evals still change verdict between two consecutive 3-sample runs (`composer_no_version_field`, `plugin_json_structure`, `release_process_steps` 0→1, `readme_requirements` 1→0), all at the 0↔1 boundary. The aggregate is steady (24 vs 26 discriminating checks, 8 vs 6 evidence-free); the per-eval membership is not.

Sampling still earns its place — it removes the single-answer coin flip and the totals are now comparable across runs — but `--require-delta` is not yet safe to gate a merge on. Filed with the numbers and three ways forward as #238.

Issue #238, opened 2026-08-19
n=3 sampling does not stabilise per-eval verdicts — four evals still flip between runs


CybotTM:
`--samples=3` (merged in [#237](https://github.com/netresearch/skill-repo-skill/pull/237), closing #236) reduces per-eval instability but does not remove it. I promised in that PR to post the two-run comparison either way; here it is, and it goes against the assumption the flag was built on.

**Two consecutive 3-sample runs, same suite, same tuple** (`skill-repo@1.35.0 · claude-code 2.1.235 · sonnet · 21 evals`), nothing changed in between:

| | run 1 | run 2 |
|---|---|---|
| baseline | 43/80 | 42/80 |
| with skill | 65/80 | 61/80 |
| discriminating checks | 24 | 26 |
| evidence-free evals | 8 | 6 |

**Four evals changed verdict between the two runs**, all at the 0↔1 boundary:

| eval | run 1 | run 2 |
|---|---|---|
| `composer_no_version_field` | 0 | 1 |
| `plugin_json_structure` | 0 | 1 |
| `release_process_steps` | 0 | 1 |
| `readme_requirements` | 1 | 0 |

That is the shape to expect: an eval whose *single* discriminating assertion is marginal flips whenever that assertion's majority does. Averaging three samples moves the coin flip one level down; it does not remove it. So `--require-delta` would still fail a different set of evals on different days — the failure mode #236 set out to fix, at a third of the amplitude and triple the cost.

**What actually survives the noise is the aggregate**: 24 vs 26 discriminating checks, 8 vs 6 evidence-free — the suite-level picture is consistent across runs even while individual membership moves.

Three ways forward, in the order I would consider them:

1. **Gate on the aggregate, not per eval.** e.g. "≥ X% of evals contribute at least one discriminating check", or a floor on total discriminating checks. Stable at N=3, no extra cost, and it matches what the numbers actually support. `skill-quality.md` already says the aggregate is what survives — the gate should follow that sentence rather than contradict it.
2. **Require K consecutive runs before calling an eval evidence-free.** Keeps the per-eval verdict but only acts on a repeated one. Costs K sweeps of latency, not tokens.
3. **Raise N.** Straightforward, and the one option whose cost is unbounded — the flips are at the boundary, so it takes a lot of samples to move a marginal assertion off it.

[...]

PR #239, opened 2026-08-19
feat(validate): catch a shebang without the committed exec bit

### LINKED ISSUE #235: A local ruff run cannot clear EXE001 — check the committed mode instead
`validate.yml` runs `uvx --no-build ruff@0.16.0 check --extend-exclude .skill-repo-tools .`, and the README tells contributors to run the linters locally before pushing. For `EXE001` (*shebang present but file is not executable*) that instruction does not hold, at least on WSL2: the same command, same pinned version, same tree, passes locally on a file the runner rejects.

Measured on 2026-08-19 while [#234](https://github.com/netresearch/skill-repo-skill/pull/234) added `scripts/no-delta-line.py`:

| | |
|---|---|
| file mode | `0o644` — `ls -l`, `stat`, `os.access(X_OK)` all agree it is not executable |
| filesystem | native ext4 (`/dev/sdd`), not a DrvFs/9p mount |
| rule enabled | yes — `EXE001` is in `linter.rules.enabled` in the local `--show-settings` dump |
| file inspected | yes — a copy with an unsorted import gets `I001` from the same run |
| local verdict | `All checks passed!` |
| runner verdict | `EXE001 Shebang is present but file is not executable → scripts/no-delta-line.py:1:1` |

So the rule is active, the file is genuinely non-executable, ruff does look at it, and the finding still does not appear locally. **The mechanism is unknown** — an earlier note in my own memory blamed DrvFs mounts reporting everything executable, and the measurements above disprove that for this machine. I am not going to substitute another guess.

What would help contributors regardless of the cause:

- **A note in the contributing/validation docs**: a local ruff run does not clear `EXE001`. Verify the committed mode instead — `git ls-tree HEAD <file>` must show `100755` for a script with a shebang.
- **Optionally a mechanical check** that needs no ruff at all and is filesystem-independent, e.g. in `validate-skill.sh`:

  ```bash
  git ls-files -s '*.py' | awk '$1=="100644"{print $4}' | while read -r f; do
    case "$(git cat-file -p "HEAD:$f" | head -1)" in '#!'*) echo "$f carries a shebang but is committed 100644";; esac
  done
  ```

  That reads the index rather than the working tree, so it gives the same answer everywhere — which is the property the ruff rule lacks here.

Filed because the finding currently lives only in a PR body and a local memory file.

CybotTM:
Closes #235.

`ruff EXE001` fails the build on a mode-`0644` Python file carrying a shebang, and on at least one developer setup it cannot be reproduced locally — same pinned ruff, same command, same file, green here and red on the runner. The mechanism is still unestablished (#235 has the measurements), so the operative problem stands regardless: a contributor's local run is not evidence, and the first signal is a red job on someone else's push.

**`validate-skill.sh` now reads the index rather than the working tree**, so the answer does not depend on what the filesystem reports:

```bash
git ls-files -s -- '*.py' '*.sh' | awk '$1=="100644"{…}'   # then check for a leading #!
```

**Severity follows what CI already enforces**, deliberately:

| | verdict | why |
|---|---|---|
| `*.py` | error | ruff fails the build on exactly these, so no repo turns newly red that was not already |
| `*.sh` | warning | nothing fails today; a shebang on a file only ever invoked as `bash <file>` is decorative, and erroring would impose a convention nobody agreed to |

That split matters for a validator consumed by ~25 repos through the reusable workflow.

**It found three files here on its first run** — `sync-plugin-manifest.sh` and two test scripts, all committed `100644` with a shebang. Fixed in the same commit (`git ls-tree` now shows `100755` for all three).

**Verified against a throwaway repo built to contain exactly the defect**: a `skills/demo/scripts/tool.py` committed `100644` with `#!/usr/bin/env python3`. The error fires and names the file. Existing suites unaffected: `tests/validate-skill.sh` 21/21, `tests/mechanical-coverage.sh` green, `pre-commit run --files <changed>` clean.

## Also in here: a caveat the second sampled run forced

[#237](https://github.com/netresearch/skill-repo-skill/pull/237) shipped `--samples=3`. I promised there to post the two-run comparison either way, and it went against the assumption: two consecutive three-sample runs of this repo's own suite still disagreed about four evals, all at the 0↔1 boundary. The aggregate held (24 vs 26 discriminating checks); the per-eval membership did not.

`skill-quality.md` now says to read a `--require-delta` failure as *"these evals were marginal in this run"* rather than *"these evals carry no evidence"*, and points at [#238](https://github.com/netresearch/skill-repo-skill/issues/238), which weighs gating on the aggregate instead. The doc would otherwise tell the next reader something this session already measured to be false.


CybotTM:
**Self-review** — no bot reviewer available (Copilot quota out for the month). Reviewed the diff myself; recording what I checked, since I also wrote it.

**No repo turns newly red.** The error tier is `*.py` only, which is exactly what `ruff EXE001` already fails the build on — a repo with that defect is red today. `*.sh` is a warning, because nothing enforces it and a shebang on a file invoked as `bash <file>` is decorative. For a validator ~25 repos consume through the reusable workflow, that split is the whole safety of this change.

**The check reads the index**, so it cannot inherit the reproducibility problem it exists to work around: `git ls-files -s` reports the committed mode regardless of what the working tree or the filesystem says.

**It can fail.** Verified against a throwaway repo containing exactly the defect — a `100644` `.py` with a shebang — where the error fires and names the file. And it found three real cases here on its first run, which are fixed in the same commit (`git ls-tree` shows `100755` for all three).

**No existing verdict moves:** `tests/validate-skill.sh` 21/21, `tests/mechanical-coverage.sh` green, `pre-commit run --files <changed>` clean, `shellcheck` clean.

**The doc change is a correction, not an addition.** `skill-quality.md` implied a sampled per-eval verdict is firm; the second sampled run measured four evals still flipping, so it now says what a `--require-delta` failure actually means and points at #238. Leaving it would have had the file assert something this session already disproved.

Checks: 22 pass, 0 fail, 2 skip (of 24). Threads: 0. `mergeState=CLEAN`. Closes #235.

Merging on that basis — reading "mergen wenn grün, und weiter" as covering the follow-up PRs of this same task.

PR #240, opened 2026-08-19
feat(evals): gate on the evidence ratio, which holds still between runs

### LINKED ISSUE #238: n=3 sampling does not stabilise per-eval verdicts — four evals still flip between runs

CybotTM:
Closes #238, taking its option 1 — gate on the aggregate — because that is the option the measurements support.

**What #238 measured.** Two consecutive `--samples=3` runs of this repo's own suite, nothing changed in between, disagreed about four evals (all at the 0↔1 boundary, where one marginal assertion decides the verdict). What held still was the aggregate: 24 vs 26 discriminating checks, 8 vs 6 evidence-free out of 21.

**`--min-evidence-ratio=P`** fails the run unless at least P percent of the **measured** evals carry a discriminating check. Unmeasured pairs are excluded from both sides of the fraction, so a limit-blocked sample cannot make a suite look worse than it is, and a run that measured nothing fails outright rather than dividing by zero.

That gives the two flags different jobs, which is now written down:

| flag | job |
|---|---|
| `--require-delta` | a **reading** signal — "these evals were marginal in this run"; open the two arms and decide. Not for blocking a merge. |
| `--min-evidence-ratio` | the **gate**, on the quantity that survives the noise. |

`evidence_ratio` also lands in `ab-results.json`, so the sweep and the dashboard can quote the stable number instead of a membership list that moves.

**Picking the floor is itself a measurement.** The docs say so explicitly: this suite came out at 62% and 71% on consecutive runs, so a 50% floor holds where 70% would fail every other time — the same coin flip one level up, which is exactly what this change exists to avoid.

**Verified:** the `mixed` fixture has a ratio of exactly 50% by construction, so both sides of the floor are testable without depending on which evals happen to be marginal — at the floor passes, below it fails, a clean suite clears a 100% floor, a fully blocked run fails, and a non-numeric floor is rejected before any model call. All 9 `tests/*.sh` suites pass; `shellcheck` and `pre-commit run --files <changed>` clean.

**Not done here, deliberately:** nothing is wired to enforce the new flag yet. The sweep still reports rather than gates, and picking a floor per repo needs a measurement per repo — that is a decision with numbers attached, not a default I should choose for 25 repos in this PR.

### DIFF
--- a/scripts/run-ab-evals.sh
+++ b/scripts/run-ab-evals.sh
@@ -1,11 +1,16 @@
 #!/bin/bash
 # run-ab-evals.sh - A/B test evals WITHOUT vs WITH skill context
-# Usage: ./scripts/run-ab-evals.sh [--no-llm] [--samples=N] [--require-delta] [concurrency]
+# Usage: ./scripts/run-ab-evals.sh [--no-llm] [--samples=N]
+#            [--min-evidence-ratio=P] [--require-delta] [concurrency]
 # Default concurrency: 4 (parallel eval pairs)
 # --no-llm: Skip LLM-based grading of expectations (regex assertions only)
 # --samples=N: completions per arm (default 1). Every per-assertion verdict is
 #   then a majority over the samples. --require-delta demands N>=2, because a
 #   verdict from a single completion flips between runs on unchanged evals.
+# --min-evidence-ratio=P: exit 1 unless at least P percent of the MEASURED evals
+#   carry a discriminating check. This is the gate the measurements support: at
+#   --samples=3 the per-eval membership still moves between runs, the ratio does
+#   not. Prefer it over --require-delta for anything that blocks a merge.
 # --require-delta: exit 1 if any eval has no discriminating assertion, i.e. no
 #   assertion that the no-skill baseline fails and the with-skill run passes.
 #   Such an eval measures boilerplate the model produces either way; it cannot
@@ -30,6 +35,11 @@ REQUIRE_DELTA=false
 # but a single sample cannot support a gate: unchanged evals were measured
 # flipping between discriminating and not on consecutive runs (issue #236).
 SAMPLES=1
+# Fraction of measured evals that must contribute at least one discriminating
+# check, as a percentage. Empty = not enforced. This is the gate that survives
+# the noise: per-eval membership moves between runs even at --samples=3, while
+# the ratio does not (issue #238).
+MIN_EVIDENCE_RATIO=""
 CONCURRENCY=4
 EVALS_FILE=""
 SKILL_FILE=""
@@ -38,6 +48,7 @@ for arg in "$@"; do
         --no-llm) NO_LLM=true ;;
         --require-delta) REQUIRE_DELTA=true ;;
         --samples=*) SAMPLES="${arg#--samples=}" ;;
+        --min-evidence-ratio=*) MIN_EVIDENCE_RATIO="${arg#--min-evidence-ratio=}" ;;
         --evals=*) EVALS_FILE="${arg#--evals=}" ;;
         --skill=*) SKILL_FILE="${arg#--skill=}" ;;
         [0-9]*) CONCURRENCY="$arg" ;;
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review, because Copilot is out of review quota for the month — no bot review will land on this or any other PR until the reset, so this is the only review the change gets. I read the diff as a reviewer rather than as its author; what follows is what that pass produced, including the one thing I would have flagged in someone else's PR.

**Correctness of the fraction.** `EVALS_UNMEASURED` and `EVALS_WITHOUT_DELTA` are disjoint by construction, not by convention: an unmeasured eval hits `continue` at `scripts/run-ab-evals.sh:458`, well before the block at `:560` that appends to `EVALS_WITHOUT_DELTA`. So `measured - without_delta` cannot double-subtract, which is the way this arithmetic would most plausibly be wrong.

**The zero case is closed on both sides.** The shell exits 1 with a named reason when `measured_evals <= 0`; the Python that writes `ab-results.json` guards the same division with `max(1, …)`, so a fully blocked run yields `evidence_ratio: 0.0` in the file instead of a traceback. Both were asserted in `tests/run-ab-evals.sh`.

**One real inconsistency, deliberately kept.** The shell computes the gating percentage with integer division while the JSON rounds to one decimal, so a suite at 66.67% prints `66%` and records `66.7`. The gate therefore truncates, which errs toward failing a borderline run rather than passing it — the safe direction for a floor. Making the two agree would mean scaled arithmetic or `awk` in the gate path for a difference that cannot change a verdict by more than one point. Worth knowing when reading the two numbers side by side; not worth the added arithmetic.

**Scope check.** The flag is opt-in and nothing sets it, so no repo's CI changes behaviour on merge. That is intentional and stated in the description: a floor is a per-repo measurement, and choosing one for 25 repos without measuring each is the mistake this whole issue is about.

`mergeState=CLEAN`, 22 checks pass, 0 fail, 0 unresolved threads.

CybotTM:
Rebased onto `main` after [#231](https://github.com/netresearch/skill-repo-skill/pull/231) landed. Head is now `6bc30c4`; the self-review above was written against `b6d6b43` and still applies to this head verbatim — `git patch-id --stable` is `4da926a…` for both commits, so the change itself is byte-identical and the only difference between the two heads is the docs file that came in from main. All nine `tests/*.sh` suites re-run green on the rebased tree, and CI is green on `6bc30c4`.

Issue #214, opened 2026-08-12
validate-skill.sh checks only the first SKILL.md, other skills in the repo go unvalidated


CybotTM:
`validate-skill.sh` stops at the first `SKILL.md` it finds. In a repo shipping several skills, only one is validated — the rest get no frontmatter check, no `Use when` check, no word count.

```bash
SKILL_FILE=""
if [[ -f "$REPO_DIR/SKILL.md" ]]; then
    SKILL_FILE="$REPO_DIR/SKILL.md"
else
    for f in "$REPO_DIR"/skills/*/SKILL.md; do
        if [[ -f "$f" ]]; then
            SKILL_FILE="$f"
            break          # <-- first match wins, the others are never seen
        fi
    done
fi
```

## Observed

`netresearch/matrix-skill` ships three skills. CI reports one line:

```
OK: SKILL.md is 476 words (under 500 limit)
```

476 is `skills/matrix-administration/SKILL.md`, alphabetically first. The actual word counts:

| file | words |
|------|-------|
| skills/matrix-administration/SKILL.md | 476 |
| skills/matrix-announcement/SKILL.md | 582 |
| skills/matrix-communication/SKILL.md | 896 |

Two of three are over the 500-word cap and CI has never said so. `matrix-communication` is the skill that gets loaded most.

The output is not misleading on its face — it names the file it checked — but "Skill Validation ✓" on a multi-skill repo reads as "the skills are valid", and the version-parity check downstream has the same single-file assumption.

## Wanted

Iterate over every `skills/*/SKILL.md` and report per skill, with the exit code covering all of them.

Fixing this will turn several repos red at once — the caps were simply never enforced there. Worth a heads-up in the release notes rather than a silent bump.

PR #243, opened 2026-08-19
fix(validate-evals): validate every evals.json, not the first one found

CybotTM:
`validate-evals.sh` had the same first-match-wins hole the skill validator had in #214: discovery stopped at the first `skills/*/evals/evals.json`, then `evals/evals.json`. A repo shipping several skills ships several eval files, and the rest were never checked.

**Measured across the 51 skill repos checked out locally — six carry more than one:**

| repo | eval files | validated before |
|---|---|---|
| `matrix-skill` | 3 | 1 (`matrix-administration`) |
| `agent-rules-skill` | 2 | 1 |
| `git-workflow-skill` | 2 | 1 |
| `go-development-skill` | 2 | 1 |
| `automated-assessment-skill` | 2 | 1 |
| `it-account-lifecycle-skill` | 2 | 1 |

All of them happen to be valid today — I ran each file individually to confirm — so nothing was silently broken. The **coverage** was, and that is the thing that fails quietly: the next defect in a non-first file would have shipped.

With more than one candidate the script now runs once per file, so each keeps its own counters and summary, and the exit code covers all of them. A single candidate, or an explicit path argument, behaves exactly as before.

## The test, and the first version of it that was worthless

Two assertions plus one guarding them:

1. the padded good fixture passes on its own,
2. a defect in a *later* file fails the run,
3. both files appear in the output, in order.

Assertion 1 exists because the first version of this test did not have it and was passing for the wrong reason: my "good" fixture failed the unified-format rules on its own (fewer than 10 evals, non-integer id, one assertion), so the run went red no matter which file was reached. Padding it through the existing `suite()` helper fixed that.

Verified against a copy of the script with first-match-wins restored: there the run goes **green with a broken second file**, and assertions 2 and 3 both fail. On this branch all three pass.

All nine `tests/*.sh` suites green, `shellcheck` and `pre-commit run --files` clean.

### DIFF
--- a/skills/skill-repo/scripts/validate-evals.sh
+++ b/skills/skill-repo/scripts/validate-evals.sh
@@ -110,12 +110,33 @@ for arg in "$@"; do
 done
 
 if [[ -z "$EVALS_FILE" ]]; then
+  # Every candidate, not the first: a repo shipping several skills ships several
+  # evals.json, and stopping at the first left the rest unchecked — six repos in
+  # the fleet carry more than one, matrix-skill three. Same hole the skill
+  # validator had (issue #214).
+  EVALS_FILES=()
   for candidate in skills/*/evals/evals.json evals/evals.json; do
-    if [[ -f "$candidate" ]]; then
-      EVALS_FILE="$candidate"
-      break
-    fi
+    [[ -f "$candidate" ]] && EVALS_FILES+=("$candidate")
   done
+
+  if [[ ${#EVALS_FILES[@]} -gt 1 ]]; then
+    # One run per file so each gets its own counters and summary, with the exit
+    # code covering all of them.
+    MULTI_RC=0
+    for candidate in "${EVALS_FILES[@]}"; do
+      echo "=============================================="
+      bash "$0" "$candidate" || MULTI_RC=1
+      echo ""
+    done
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review — Copilot has no review quota left this month.

The change is a wrapper around unchanged logic: with more than one candidate the script re-invokes itself per file and aggregates the exit code; with one candidate or an explicit path it takes the same branch it always did. That containment is the reason I did not restructure the 350 lines of checks into a function.

**The test is where this review has something to say.** My first version passed for the wrong reason: the "good" fixture failed the unified-format rules on its own — fewer than 10 evals, a non-integer id, one assertion — so the run went red whichever file was reached, and the multi-file assertion could not fail. It now goes through the existing `suite()` helper, which pads the file, and an explicit assertion pins that the good fixture passes alone. Without that guard the other two assertions are decoration.

Verified against a copy of the script with first-match-wins restored: the run goes **green with a broken second file**, and both new assertions fail there. On this branch all three pass.

**Fleet check before claiming the gap mattered**: six repos carry more than one `evals.json`, and I ran each file individually — all valid today. So nothing was silently broken; the coverage was, which is the failure mode that only shows up when someone finally introduces a defect in a non-first file.

22 checks pass, 0 unresolved threads, `mergeState=CLEAN`.

PR #244, opened 2026-08-19
feat(tests): run PHP test entrypoints, not only shell and Python

### LINKED ISSUE #242: tests.yml has no PHP leg, so the "scripts are tested" rule has no gate for repos that ship PHP
### Observed friction

`repository-quality-rules.md` states that a script shipped under `scripts/` must be referenced by a test under `tests/`, and names `netresearch/skill-repo-skill/.github/workflows/tests.yml@main` as the runner that makes it a gate rather than a convention. That reusable globs `tests/**/*.sh` and `tests/**/*.py` only. A skill repo whose tests are PHP therefore satisfies the rule on paper while no pipeline ever executes them.

This surfaced while bootstrapping [netresearch/php-ast-edit-skill](https://github.com/netresearch/php-ast-edit-skill), the first skill repo that is also a runnable PHP package (`src/`, `bin/`, PSR-4, `nikic/php-parser`). Its real test needs `composer install` and a PHP version matrix, neither of which the reusable provides.

### Desired future behavior

A skill repo that ships code in a language outside the shell/Python pair either gets a runnable leg in the shared reusable, or the standard states plainly that such a repo owns a local workflow and what that exception must look like. Today an author has to discover the gap by reading the reusable.

### Proposed skill change

Two routes, and the choice is a maintainer call rather than something a single consuming repo should decide:

1. Add an optional PHP leg to `tests.yml` (a `php_glob` input plus a `composer install` step, mirroring the existing `shell_glob` / `python_glob` shape), so the gate covers the language the fleet actually writes most of.
2. Or document the exception in `repository-quality-rules.md`: a repo shipping non-shell/Python code declares a repo-local workflow and records it under `## Repository extras` in its README.

php-ast-edit-skill currently bridges it both ways as a stopgap: a `tests/run.sh` wrapper that calls `php scripts/check.php`, `php tests/run.php`, the skill wrapper and the PHAR build (so the shell glob has something to find), plus a repo-local `php-tests.yml` running the real matrix on PHP 8.2/8.3/8.4. The exception is documented in its README. That combination works but every future PHP-carrying skill repo will re-derive it.

### Target area

`.github/workflows/tests.yml` and `skills/skill-repo/references/repository-quality-rules.md`.

### Origin session

Found while creating php-ast-edit-skill; related documentation change merged as #231.

CybotTM:
Closes #242.

`repository-quality-rules.md` names this workflow as the thing that turns "every script under `scripts/` is referenced by a test" into a gate rather than a convention. The workflow globbed `tests/**/*.sh` and `tests/**/*.py` only, so a skill repo whose tests are PHP satisfied the rule on paper while no pipeline ever executed them.

That is not hypothetical: [php-ast-edit-skill](https://github.com/netresearch/php-ast-edit-skill) is the first skill repo that is also a runnable PHP package, and it currently bridges the gap with a `tests/run.sh` wrapper whose only job is to give the shell glob something to find, plus a repo-local matrix workflow. Once this lands it can call the reusable directly.

## What the leg does

A detection step resolves `php_glob` in plain bash, and `shivammathur/setup-php` plus `composer install` run only when it matched something — a repo shipping no PHP pays a few milliseconds of globbing and nothing else. `php_version` defaults to 8.3 and is an input; a repo wanting a real matrix keeps that in its own workflow, since a reusable running one version is a gate, not a compatibility matrix.

`composer install` runs with `--no-plugins`, deliberately. Every skill repo requires `netresearch/composer-agent-skill-plugin`, and a root install that has not declared it under `config.allow-plugins` aborts non-interactively — the trap documented in `composer-setup.md` since #231. Running tests never needs that plugin, so the gate stays independent of each repo's allow-plugins state and adoption cannot break on it.

The coverage tally at the end now counts PHP files too, otherwise a PHP-only repo would still be reported as shipping scripts with no tests.

`php_glob` matches are all executed, so its description says plainly that a repo keeping fixtures under `tests/` narrows the glob to its entrypoints. php-ast-edit-skill would pass `tests/run.php`, not the default, because the default would also execute `tests/fixtures/sample.php`.

## Also in this diff

`docs/ARCHITECTURE.md` was missing `tests.yml` from the reusable inventory entirely — pre-existing drift, not something this change introduced. Since the AGENTS.md rule binds those documents as a set and this PR modifies the workflow, the entry is added here rather than left for a follow-up.

## Verification

`actionlint` and the repo's pre-commit suite pass; `validate-skill.sh` reports 0 errors, 0 warnings. What that does not show is the leg actually running, because no repo calls it yet. An end-to-end run against php-ast-edit-skill is the real proof and is linked in a comment below once it exists.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,6 +23,20 @@ on:
         required: false
         type: string
         default: "tests/**/*.py"
+      php_glob:
+        description: >-
+          Glob for PHP test entrypoints, relative to the repo root. Every match
+          is executed, so a repo keeping fixtures under tests/ narrows this to
+          the entrypoints (for example "tests/run.php") rather than letting the
+          default sweep tests/fixtures/ as well.
+        required: false
+        type: string
+        default: "tests/**/*.php"
+      php_version:
+        description: "PHP version used when PHP tests are present."
+        required: false
+        type: string
+        default: "8.3"
       require_tests:
         description: >-
           Fail when the repo ships scripts under skills/*/scripts/ but matches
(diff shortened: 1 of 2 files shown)


CybotTM:
End-to-end verification, since the diff alone cannot show a workflow running. A caller in [php-ast-edit-skill](https://github.com/netresearch/php-ast-edit-skill) pinned to this branch with `php_glob: tests/run.php`.

**First run ([32269705621](https://github.com/netresearch/php-ast-edit-skill/actions/runs/32269705621)) — green, and that was the problem.** All four PHP steps executed rather than skipping (`Detect PHP tests`, `Set up PHP`, `Install PHP dependencies`, `PHP tests`), but the log showed the shell leg reporting success while doing almost nothing:

```
PHP syntax OK (9 files)
SKIP: vendor/autoload.php missing; run composer install to execute integration tests.
SKIP: vendor/autoload.php missing; wrapper needs the parser.
SKIP: vendor/autoload.php missing; PHAR build needs the parser.
```

`tests/run.sh` runs before `composer install`, so three of its four checks skipped themselves and only `php -l` actually ran. The suite passed for real afterwards in the PHP leg, but a repo relying on the shell wrapper alone — which is what php-ast-edit-skill does today — has a gate that reports green while guarding nothing.

**Second commit moves the toolchain ahead of every leg.** Same branch, same tests, only the ordering changed:

```
- Installing nikic/php-parser (v5.8.0): Extracting archive
PHP syntax OK (9 files)
OK
/home/runner/work/php-ast-edit-skill/php-ast-edit-skill/dist/php-ast-edit.phar
OK: all checks passed.
OK
shipped scripts: 1 | test files run: 2
```

Zero `SKIP:` lines in [32270100938](https://github.com/netresearch/php-ast-edit-skill/actions/runs/32270100938), down from three. The PHAR build and the wrapper check now execute, and `tests/run.php` runs the inspect/apply round-trip in both legs.

The detection step is still pure bash and runs first, so a repo shipping no PHP gets one glob expansion and nothing else.

The caller branch is temporary and pinned to `@feat/tests-php-leg`; it gets repointed at `@main` and proposed as php-ast-edit-skill's adoption once this merges.

CybotTM:
Third end-to-end run after the review fix ([32272253050](https://github.com/netresearch/php-ast-edit-skill/actions/runs/32272253050)): all seven steps green, zero `SKIP: vendor` lines, `OK: all checks passed.`, PHAR built, `shipped scripts: 1 | test files run: 2`.

The fix itself came out of reviewing my own reorder: `Shell tests` carried no `if:` because it used to be the first step. With the toolchain moved ahead of it, a failed `composer install` would have skipped the shell leg while Python and PHP still ran on `always()` — the "one kind of failure hides the others" shape this workflow argues against two comments further down. It now has `always()` like its siblings. The unused `id: php` is gone; the tally reads the detection step.

PR #246, opened 2026-08-19
fix: add apt timeouts and job timeout to prevent indefinite hangs (#245)

### LINKED ISSUE #245: validate.yml can hang indefinitely: apt-get has no acquisition timeout and its output is discarded
`validate.yml`'s ShellCheck step installs its tool with

```
sudo apt-get update -qq > /dev/null 2>&1 && sudo apt-get install -y -qq shellcheck > /dev/null 2>&1
```

apt applies no timeout to package acquisition by default, so a mirror that accepts the connection and then stops responding blocks forever rather than failing. On 2026-08-19 that stalled the step for 37 minutes in [netresearch/php-ast-edit-skill](https://github.com/netresearch/php-ast-edit-skill) ([job 96132793056](https://github.com/netresearch/php-ast-edit-skill/actions/runs/32272728286/job/96132793056), started `15:54:48Z`) until it was cancelled by hand. A re-run of the same commit passed.

The same signature hung an unrelated workflow in another repository the same afternoon — `playwright install --with-deps`, which shells out to `apt-get update`, stalled 20 minutes on its first mirror fetch (netresearch/claude-code-marketplace#99). Two repositories, two tools, one shared dependency.

## Why it is invisible

`-qq` plus `> /dev/null 2>&1` discards everything apt would say, so a stalled fetch looks exactly like a step doing nothing. In the marketplace case the diagnosis took three runs precisely because no output survived.

## Suggested change

Two independent bounds, mirroring netresearch/claude-code-marketplace#102:

1. Write `/etc/apt/apt.conf.d` timeouts (`Acquire::Retries`, `Acquire::http::Timeout`, `Acquire::https::Timeout`) before the install so apt fails instead of hanging.
2. Wrap the install in `timeout` with a small retry loop, and stop discarding its output so the next stall is diagnosable from the log alone.

`validate.yml` has no `timeout-minutes` on its job either, so a stall there holds a runner until the six-hour ceiling. Worth adding in the same change; the other reusables in this repo are exposed the same way.

## Blast radius

This reusable is called by every skill repo in the fleet, so a bad apt day stalls validation everywhere at once rather than in one repository.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

aditya226-sharma:
## Summary
Adds apt acquisition timeouts and a job-level timeout to prevent `validate.yml` from hanging indefinitely when a mirror is slow or unresponsive.

## Changes
- **APT timeouts**: Writes `/etc/apt/apt.conf.d/99-timeouts` with `Acquire::Retries 3`, `Acquire::http::Timeout 15`, `Acquire::https::Timeout 15` before apt-get update/install
- **Install timeout**: Wraps both `apt-get update` and `apt-get install` in `timeout 60` so the step fails instead of hanging
- **Job timeout**: Adds `timeout-minutes: 15` to the validate job to cap total runner time
- **Visible output**: Removes `> /dev/null 2>&1` from apt commands so stalls are diagnosable from logs

## Why
Without these bounds, a mirror that accepts the connection and then stops responding blocks the step forever. The `-qq` plus `> /dev/null 2>&1` discards all output, making a stalled fetch invisible. On 2026-08-19 this stalled a step for 37 minutes until cancelled by hand. This reusable is called by every skill repo in the fleet, so a bad apt day stalls validation everywhere.

Closes #245

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -10,6 +10,7 @@ jobs:
   validate:
     name: Skill Validation
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - name: Harden Runner
         uses: step-security/harden-runner@05e31511f85b41b11d1cf0ef85d0992719546e2c # v2.21.0
@@ -192,7 +193,13 @@ jobs:
           fi
           COUNT=$(echo "$SCRIPTS" | wc -l)
           echo "Found $COUNT shell script(s)"
-          sudo apt-get update -qq > /dev/null 2>&1 && sudo apt-get install -y -qq shellcheck > /dev/null 2>&1
+          # Set apt timeouts to prevent indefinite hangs on slow mirrors
+          sudo mkdir -p /etc/apt/apt.conf.d
+          echo 'Acquire::Retries "3";' | sudo tee /etc/apt/apt.conf.d/99-timeouts > /dev/null
+          echo 'Acquire::http::Timeout "15";' | sudo tee -a /etc/apt/apt.conf.d/99-timeouts > /dev/null
+          echo 'Acquire::https::Timeout "15";' | sudo tee -a /etc/apt/apt.conf.d/99-timeouts > /dev/null
+          timeout 60 sudo apt-get update -qq 2>&1 | tail -5 || echo "apt-get update completed or timed out"
+          timeout 60 sudo apt-get install -y -qq shellcheck 2>&1 | tail -5 || { echo "::error::Failed to install shellcheck within timeout"; exit 1; }
           FAILED=0
           while IFS= read -r script; do
             [[ -z "$script" ]] && continue


CybotTM:
Thank you for this, and sorry it took so long to get a proper look at it.

Closing it as superseded rather than rejected: the hazard you fixed is real and is now bounded, but by a different change that landed on `main` in the meantime.

**What happened to the apt path.** `validate.yml` no longer installs shellcheck through apt at all. It downloads a pinned release tarball with `curl -fsSL --max-time 120`, verifies it against a SHA256, and installs that (`65f90ed`, `d23a978`). The step your diff patches does not exist any more, which is also why this PR now shows as conflicting.

**And the job timeout.** `timeout-minutes: 20` is on the `validate` job on `main` today. Your `15` would lower it, which is a separate judgement rather than a gap being closed.

**One thing worth passing on anyway**, because it would have bitten regardless of the above:

```bash
timeout 60 sudo apt-get install -y -qq shellcheck 2>&1 | tail -5 || { echo "::error::..."; exit 1; }
```

The step sets no `shell:` and the workflow has no `defaults`, so GitHub runs it as `bash -e {0}` — **without `pipefail`**. The pipeline's exit status is therefore `tail`'s, which is always 0, so the `|| { … exit 1; }` can never fire. A timed-out or failed install would have been swallowed and the step would have carried on to the shellcheck loop. `tail` also buffers, so during an actual stall nothing prints at all — the opposite of the visibility the change was after. Adding `shell: bash` (which turns on `pipefail`) or dropping the pipe would fix it.

**If you would like to keep going on this:** there is one genuine remaining gap in the same family. Across all 23 workflows in this repository, exactly one job has no `timeout-minutes` — `self-test` in `.github/workflows/self-test.yml`. That one is still open and would be a welcome PR.

Thanks again for the diagnosis in #245's neighbourhood — the reasoning was right, the fix just arrived from another direction first.

PR #249, opened 2026-08-20
fix(ci): make the shellcheck hook and CI the same check

### LINKED ISSUE #248: shellcheck hook and CI disagree on severity — the stricter gate is the unenforceable one
The `.pre-commit-config.yaml` this skill hands out opens with "Every hook below ALSO runs in CI (via the reusable `validate.yml` workflow from netresearch/skill-repo-skill)". For shellcheck the two do not run the same check:

- **Hook**: `shellcheck-py` with no `args`, i.e. shellcheck's default severity — `style`, which includes `info` and `warning`.
- **CI** (`validate.yml`): `shellcheck -x -S error "$script"`.

A shell asset that CI passes therefore cannot be committed locally. Observed in `netresearch/typo3-testing-skill` on a three-line change to `skills/typo3-testing/assets/Build/Scripts/runTests.sh` ([#133](https://github.com/netresearch/typo3-testing-skill/pull/133)): the hook reported 105 findings, all pre-existing and byte-identical to `origin/main`, while `shellcheck -x -S error` on the same file exits 0. The commit only went through with `SKIP=shellcheck`.

Most of those findings are `SC2086` on deliberate word splitting (`${CONTAINER_COMMON_PARAMS}` expanding into several docker arguments). "Fixing" them by quoting would break the script, so the hook is asking for something that must not be done — which is how a bypass becomes routine, and the config itself warns against exactly that: "If you need this often, the hook is wrong — fix it; don't tolerate the bypass."

Two ways out, both cheap:

1. Give the hook CI's arguments, so parity is real: `args: [-x, -S, error]`.
2. Or raise CI to the hook's severity and annotate the intentional findings in the assets with `# shellcheck disable=SC2086` where splitting is meant.

(1) restores the stated invariant without touching any asset. (2) is more thorough and costs a pass over every shipped script. Either is better than the current state, where the stricter gate is the one that cannot be enforced.

Whichever is chosen, the `-x` flag matters: CI passes it, the hook does not, so a script with `source`d files is analysed differently in the two places even at equal severity.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

CybotTM:
Closes #248.

The config header promises "every hook below ALSO runs in CI". For shellcheck it did not:

| | hook | `validate.yml` |
|---|---|---|
| severity | shellcheck's default (`style`) | `-S error` |
| `-x` (follow `source`d files) | no | yes |

So a script CI accepted could be uncommittable, and the only way past it was `SKIP=shellcheck` — in a config that explicitly warns "if you need this often, the hook is wrong; don't tolerate the bypass".

## Not lowering one side blindly

The obvious fix — give the hook `-S error` — weakens a gate that most repositories pass today. Measured: all 21 shell scripts in this repository are clean at `style`. So the severity becomes an **input** instead, defaulting to `error`, which leaves all 23 consumers of `validate.yml` exactly where they are:

```yaml
uses: netresearch/skill-repo-skill/.github/workflows/validate.yml@main
with:
  shellcheck-severity: style   # only where the tree is clean at that level
```

A repository then pins the same level on **both** sides. `style` where its scripts are its own; `error` where it carries a vendored script it does not own — `netresearch/typo3-testing-skill` ships a TYPO3-Core-derived `runTests.sh` whose 105 findings are mostly deliberate word splitting (`${CONTAINER_COMMON_PARAMS}` must expand into several docker arguments; quoting it would break the script).

## This repository takes the strict side

`self-test.yml` now calls the validator with `shellcheck-severity: style`, and the hook gains `-x` to match CI. That call closes a second hole while it is there: `lint.yml` calls `validate.yml@main`, so a change to the validator was never gated by the pull request making it — this one is.

`lint.yml` deliberately keeps calling `@main` **without** the new input: passing an input that `main` does not declare yet fails the run at startup. It follows in a one-line PR once this is merged.

## Verified

- `shellcheck -x -S style` over all 21 scripts: 0 findings.
- `tests/validate-skill.sh`, `tests/portable-manifest.sh`, `validate-skill.sh .`, `check-version-parity.sh`: all exit 0.
- `actionlint -shellcheck=shellcheck` on both changed workflows: clean.
- The new input is exercised by this PR's own run, not only after merge — that is what the `validate-self` job is for.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

### DIFF
--- a/.github/workflows/self-test.yml
+++ b/.github/workflows/self-test.yml
@@ -43,3 +43,16 @@ jobs:
 
       - name: Version parity
         run: bash skills/skill-repo/scripts/check-version-parity.sh
+
+  # The shellcheck step lives in validate.yml, and lint.yml calls that file from
+  # `main` — so a change to it is not gated by the PR that makes it. Calling the
+  # PR's own copy closes that hole, and pins this repo to the severity its
+  # pre-commit hook already enforces: all 21 scripts here are clean at `style`,
+  # so hook and CI agree by construction instead of by claim.
+  validate-self:
+    name: Skill Validation (this PR's validator)
+    uses: ./.github/workflows/validate.yml
+    permissions:
+      contents: read
+    with:
+      shellcheck-severity: style
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -221,7 +221,11 @@ jobs:
           # the same file, which is the divergence this input exists to end.
           # It also removes the apt stall that #245 had to bound: a pinned
           # tarball with a checksum cannot hang for 37 minutes on a mirror.
-          curl -fsSL --max-time 120 -o /tmp/shellcheck.tar.xz \
+          # --proto/--proto-redir pin the scheme for the request AND for the
+          # redirect GitHub answers with, so -L cannot be walked onto http
+          # (githubactions:S6506).
+          curl --proto '=https' --proto-redir '=https' --tlsv1.2 \
+            -fsSL --max-time 120 -o /tmp/shellcheck.tar.xz \
             "https://github.com/koalaman/shellcheck/releases/download/v${SHELLCHECK_VERSION}/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz"
           echo "${SHELLCHECK_SHA256}  /tmp/shellcheck.tar.xz" | sha256sum -c
           tar -xJf /tmp/shellcheck.tar.xz -C /tmp "shellcheck-v${SHELLCHECK_VERSION}/shellcheck"
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -2,6 +2,21 @@ name: Validate Skill
 
 on:
   workflow_call:
+    inputs:
+      shellcheck-severity:
+        description: >-
+          Minimum shellcheck severity that fails the job: error (default),
+          warning, info or style. It exists so a repository can hold its
+          pre-commit hook and this job to the SAME level — the hook runs
+          shellcheck at its own default (style) unless given `args`, and a
+          repository whose scripts are clean at that level should say so here
+          rather than let the two gates disagree. Raise it only once the tree
+          passes: a repository carrying a vendored script it does not own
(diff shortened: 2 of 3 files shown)


CybotTM:
**Review note.** The `copilot_code_review` ruleset could not deliver a review — the account's monthly Copilot quota is exhausted, so the bot returns an error instead of a review and re-requesting cannot change that before the reset. Reviewed by hand; recording what was checked.

- The first plan for this issue was to give the hook CI's `-S error`. Measuring killed it: all 21 shell scripts here pass at `style`, so lowering the hook would have weakened a gate that works. The severity became an input instead, defaulting to `error`, so none of the 23 consumers of `validate.yml` changes behaviour.
- The self-test job added in the first commit **failed on its first run**, which is the only reason the version divergence was found: CI's apt shellcheck is 0.9.0, the hook pins 0.11.0.1, and 0.9.0 reports SC2317 where 0.11.0 does not. Confirmed by running `apt-get install shellcheck` in an `ubuntu:24.04` container and reading `shellcheck --version`, not by assuming what Ubuntu ships.
- The pinned download replaces an apt install that #245 had to bound against a 37-minute stall. Both the URL and the checksum were exercised locally, including after the `--proto`/`--proto-redir` hardening SonarCloud asked for — same sha256 either way.
- One trap in the change itself: a comment beginning `# shellcheck ` **is** a shellcheck directive, so the first version of that comment failed actionlint. Reworded rather than annotated away.
- The two version pins — `rev:` in `.pre-commit-config.yaml` and `SHELLCHECK_VERSION` in `validate.yml` — are one pair, and the config now says so; bumping one alone puts the gates back into disagreement.
- Not done here: `lint.yml` still calls `validate.yml@main` without the new input, because passing an input `main` does not declare yet fails the run at startup. It follows in a one-line pull request once this is merged.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01GSptxPLHWsttu9FuqVkvYZ)_

CybotTM:
Self-review: d23a9786e1220a4e023c8951616ea0a3826b0950

PR #250, opened 2026-08-21
feat(validate): enforce the Agent Skills budgets and flat discovery

CybotTM:
Enshrines the Agent Skills architecture rules as the base for every skill repo: **flat discovery, small routing metadata, small control plane, details on demand.**

## What was wrong

**The size budget measured the wrong thing.** The check counted **500 words over the whole file**. The spec says *"Keep your main `SKILL.md` under 500 lines"* — words are a different and far tighter constraint, and counting the frontmatter charged the `description` to the body's allowance. That trade is always the wrong way round: the description is the only surface loaded at startup, so it decides whether the skill is ever activated. It has already cost real capabilities — a consuming repo sat at **499 of 500** and left a script out of its `SKILL.md` rather than spend fourteen words naming it.

**Nothing checked reachability at all.** The spec is explicit: *"Keep file references one level deep from `SKILL.md`. Avoid deeply nested reference chains."*

## Budgets now enforced, all from the spec

| Check | Level | Source |
|---|---|---|
| `description` > 1024 chars | ERROR | spec hard limit |
| `description` > 500 chars | WARN | guidance: "a few sentences to a short paragraph" |
| `name` with leading/trailing or doubled hyphen | ERROR | spec — the character class alone accepted all three |
| `compatibility` > 500 chars | ERROR | spec |
| body > 500 lines | ERROR | spec recommendation |
| body > 300 lines | WARN | ask what is control flow and what is reference material |

On the 500-character warning: it is a **nudge, not a defect**. The same guidance says to *"err on the side of being pushy"* about listing contexts, so long is not automatically wrong — long **and** narrating a workflow is. A description carrying a condensed workflow invites the agent to follow *that summary* instead of activating the skill and reading the real instructions.

## Flat discovery, now linted

Each hop is a decision the agent may not make. `SKILL.md` is read in full on activation; a reference is read only if `SKILL.md` said what it holds and when to open it. A file reachable only through a second hop sits behind a door with no sign on it.

Warnings, not errors — a chain is a smell, not a breach:

- a `references/*.md` reachable only through another reference, **naming both** so the fix is obvious
- a `references/*.md` nothing names at all
- a reference over 100 lines with no `## Contents` — agents preview long files rather than reading them whole
- an **executable** under `scripts/` that `SKILL.md` never names

Non-executable files under `scripts/` are exempt from the last one: a sourced library is not a capability the agent invokes, and its caller is what belongs in the control plane.

## Deliberately not linted

Whether a description narrates a workflow, whether a verification step exists, whether an optional frontmatter field has a consumer. No mechanical check decides those honestly, and a rule that guesses is worse than a review question. `references/skill-architecture.md` says so explicitly rather than leaving the omission to be rediscovered.

## The reference

`references/skill-architecture.md` records the whole thing with sources: the three surfaces and when each enters context, every budget and where its number comes from, the flat-discovery rule with worked good/bad layouts (including why an index file that only points at scripts should be deleted rather than kept), what belongs in `SKILL.md` versus `references/` versus `scripts/`, and the trigger-eval method — 20 queries, **near-miss** negatives, 3 runs each, train/validation split — which settles description arguments better than counting characters ever will.

It closes on a caution worth keeping: public skill repositories are **data, not best practice**. "Other people do it this way" carries no weight here.

## This repository broke four of its own new rules

Fixed in the same PR, because a linter that leaves nine warnings on its own repo is a claim rather than a rule:

- eight references over 100 lines gained a `## Contents`
- seven executables in `scripts/` are now named in `SKILL.md`
- the budget line in `SKILL.md` itself quoted the wrong numbers (`≤500 words`, `≤1,536 chars`)
- `SKILL.md` is now **0 errors, 0 warnings** under the new rules

## One change that was necessary to land this at all

The repository pinned **its own validator** through pre-commit at `v1.22.0`, so the gate on every commit was the last release rather than the working copy. Two consequences: a change to the validator could never be exercised before it shipped, and a `SKILL.md` written against the corrected budgets was rejected by the obsolete ones — with no way out but weakening the content or bypassing the gate, and `main` would have gone red after merge either way.

`validate-skill` now runs as a `repo: local` hook against the working copy. `check-version-parity` stays pinned; it is not authored against itself.

## Tests

`tests/validate-skill-architecture.sh` — 19 assertions, **13 of which fail against the previous validator**. Covers each budget at its boundary, both spec name rules, the second-hop chain, the orphan reference, the missing `## Contents`, the unnamed executable, and the library exemption that keeps that last warning honest.

The existing smoke test's multi-skill fixture moves from 520 words to 520 lines — same intent (*a violation in a later skill must still be reported*), only the metric it violates. All 10 files in `tests/` pass; `shellcheck` clean; the CI shell glob is `tests/**/*.sh`, so the new file is picked up without a workflow change.

## A bug found while wiring this up

Worth naming because it is silent: in one draft the `skill_dir` assignment sat **above** its `local` declaration, so `local` reset it, `set -u` aborted the run mid-way, and **no summary printed at all** — which reads exactly like a repository that passed.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,6 +15,12 @@ permissions: {}
 jobs:
   validate:
     name: Skill Validation
-    uses: netresearch/skill-repo-skill/.github/workflows/validate.yml@main
+    # Local reference, not netresearch/skill-repo-skill/...@main. This IS that
+    # repository: pointing at main runs the workflow this pull request is
+    # changing, from before the change, so a workflow or validator edit is
+    # judged by the version it replaces and can never be green on its own pull
+    # request. self-test.yml already calls it locally; this is the same call
+    # made consistent. Consumer repos keep the remote @main reference.
+    uses: ./.github/workflows/validate.yml
     permissions:
       contents: read
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,12 +15,6 @@ permissions: {}
 jobs:
   validate:
     name: Skill Validation
-    # Local reference, not netresearch/skill-repo-skill/...@main. This IS that
-    # repository: pointing at main runs the workflow this pull request is
-    # changing, from before the change, so a workflow or validator edit is
-    # judged by the version it replaces and can never be green on its own pull
-    # request. self-test.yml already calls it locally; this is the same call
-    # made consistent. Consumer repos keep the remote @main reference.
-    uses: ./.github/workflows/validate.yml
+    uses: netresearch/skill-repo-skill/.github/workflows/validate.yml@main
     permissions:
       contents: read
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -60,7 +60,21 @@ jobs:
           cache-dependency-glob: ""
 
       - name: Run skill validation
-        run: bash .skill-repo-tools/skills/skill-repo/scripts/validate-skill.sh .
+        # When the CALLER is skill-repo-skill itself, validate with the copy
+        # under test rather than with the one on main. Both jobs that run this
+        # workflow used the .skill-repo-tools checkout, so a pull request that
+        # changed the validator was judged by the version it was replacing --
+        # including the job literally named "this PR's validator". A validator
+        # change could therefore never be green on its own pull request.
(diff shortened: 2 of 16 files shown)


CybotTM:
Self-review: 74c080ce800170c204976db4d805b616c3cb1885

PR #264, opened 2026-08-28
fix(fleet-release): retry the bump commit once when a hook rewrites it

### LINKED ISSUE #263: sync-plugin-manifest.sh key order fights the pretty-format-json hook and aborts the fleet bump commit
`sync-plugin-manifest.sh` writes `.claude-plugin/plugin.json` with the key order of the source `plugin.json` (name, version, description, author, repository, license, keywords, skills). Repos whose `.pre-commit-config.yaml` runs `pretty-format-json` with `--autofix` sort those keys alphabetically. The committed file therefore carries the sorted order, the generator re-emits the unsorted one on every bump, and the hook rewrites it during `git commit` — which aborts the commit, because a hook that modifies files fails the run.

For a hand bump that is one re-`git add` away. For `fleet-release-common.sh` it is a hard stop: `fr_commit_allowlisted` reports `COMMIT EXIT: 1` and the repo comes out of the `bump` phase as `FAIL <repo>: commit` with a half-written worktree that `fr_resume_worktree` then refuses as dirty. The operator has to commit by hand in the release worktree and re-run `bump` to get the MR opened.

Measured in the 2026-08-28 GitLab sweep: `coding-ai/it-maintenance-skill` (v1.15.0) and `coding-ai/netresearch-jira-skill` (v2.11.1) both failed this way; the other two repos in the same sweep have no `pretty-format-json` hook and passed. The resulting trees are correct — only the driver's commit step fails.

Fixes worth considering, in order of preference:

1. `sync-plugin-manifest.sh` emits the projected manifest with sorted keys (`jq -S`), so generator and hook agree.
2. `fr_commit_allowlisted` retries the commit once after re-staging the allowlisted paths, which is the documented remedy for a reformatting hook.

CybotTM:
Fixes #263

A pre-commit hook that reformats a file it checks — `pretty-format-json --autofix`, `black`, `ruff format` — modifies the staged tree and then *fails* the commit. `fr_commit_allowlisted` treated that as final, so the repo left the `bump` phase as `FAIL <repo>: commit` with a half-written worktree that `fr_resume_worktree` then refuses as dirty; the operator had to commit by hand in the release worktree and re-run `bump` to get the PR opened. `it-maintenance-skill` v1.15.0 and `netresearch-jira-skill` v2.11.1 both went that way in the 2026-08-28 sweep.

The commit now makes two attempts and logs both (`COMMIT EXIT (1): 1`, `COMMIT EXIT (2): 0`). Staging moved into its own `fr_stage_allowlisted`, so the retry re-walks the porcelain rather than re-adding the first pass's paths — a hook that writes *outside* the version surfaces still fails the allowlist on the second pass, and a deterministic rejection still fails after exactly two attempts.

## Why not fix the generator instead

The issue offered sorting `sync-plugin-manifest.sh`'s output as the first option. It is neither necessary nor sufficient. `pretty-format-json` also escapes non-ASCII by default, and the fleet's manifest descriptions are full of em dashes and umlauts, so one edit to any hook-carrying repo's description reopens the same bounce; meanwhile the sort would reorder the generated manifest in every repo, hook or not, to avoid one hook re-run in the four that have it. Measured on the current fleet: four repos run `pretty-format-json` (`it-account-lifecycle`, `it-maintenance`, `it-offboarding`, `netresearch-jira`), all four ASCII-only today. The retry is blind to *what* the hook changed, which is the property the fix needs.

## Tests

`fr_commit_allowlisted` had no tests; these are the first, and each was observed to fail against the single-attempt engine (`for attempt in 1`):

- reformatting hook on an allowlisted file → attempt 1 fails, attempt 2 commits, worktree clean afterwards
- hook writing a non-allowlisted file → still `unexpected change outside the version surfaces`, still reported to the caller
- hook that always rejects → exactly two attempts, then `FAIL`

The fixture is a plain `.git/hooks/pre-commit` that rewrites a staged file and exits 1 once — what `--autofix` does — with an ed25519 key for the `-S` in the commit. All ten `tests/*.sh` pass, shellcheck is clean, and `references/release-discipline.md` gains the section with the two-attempt log shape.

## Follow-up

`fleet-release-gitlab.sh` in `coding-ai/gitlab-skill` runs a vendored copy of this engine, and that is where the failure actually fired. Re-vendoring is a separate MR on that repo once this lands.

Copilot review is out of quota for the month, so this merges on a documented `Self-review:` attestation.

### DIFF
--- a/skills/skill-repo/references/release-discipline.md
+++ b/skills/skill-repo/references/release-discipline.md
@@ -6,6 +6,7 @@
 - Canonical Order: Bump PR Merged → Tag Pushed
 - Pre-Release Version-Parity Check
 - Changelog Rollover in the Bump Commit
+- A reformatting hook aborts the first bump commit
 - Cache Safety: Never Edit the Installed Copy
 - Multi-Skill-Repo Release Dry-Run
 - GitLab (`git.netresearch.de`) skill repos release on tag too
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: 30387ab9ad3c2ed5679bad30754d9e91de9775df

Issue #267, opened 2026-08-28
Promote the dangling-SKILL.md-path check from WARN to ERROR once the 8 known repos are clean


CybotTM:
#266 adds a WARN for skill-relative paths that `SKILL.md` names but that do not exist under the skill directory (`scripts/`, `references/`, `assets/`, `evals/`, with or without `${CLAUDE_SKILL_DIR}/`). It is a warning only because the validator runs in every consuming repository's CI and, measured over the 120 skills installed on one machine on 2026-08-28, 8 currently carry such a path:

| Plugin / skill | Dangling path(s) |
|---|---|
| `agent-harness` | `references/agents-md-rules.md`, `references/artefact-inventory.md` |
| `automated-assessment` / `add-checkpoints` | `references/checkpoints-schema.md`, `references/migration-guide.md`, `scripts/validate-checkpoints.sh` |
| `dxp-frontend` / `dxp-frontend-audit` | `references/gitlab-ci.md`, `references/renovate.md` |
| `dxp-frontend` / `dxp-frontend-init` | `references/gitlab-ci.md`, `references/linter-configs.md`, `references/playwright.md` |
| `netresearch-account-lifecycle` / `netresearch-offboarding` | `references/_global/task-checklists.md` |
| `netresearch-branding` | `assets/logos/netresearch-symbol-only.svg` |
| `skill-repo` | `scripts/audit-skills.sh` (fixed in #266) |
| `typo3-upgrade-estimator` | `references/relaunch-vs-portation.md` |

Measured against installed cache versions; a repository may already be clean on `main`.

**Done when:** each listed repository either ships the file or corrects the path (their CI shows the WARN line after the next skill-repo-skill release), and the check in `validate-skill.sh` is switched from `warning` to `error` with the `skill-architecture.md` table row updated to ERROR.

_Assisted by claude-code:claude-fable-5 — [Session](https://claude.ai/code/session_01F91Rh11LCehJojefpo5aCF)_

CybotTM:
Fleet measured with the WARN build from #266 on 2026-08-29, this time over the whole population rather than one machine's installed plugins: the 40 GitHub repositories in `fleet-repos-github.txt` and the 37 `coding-ai` GitLab projects that ship `skills/`. Findings and fixes:

| Repository | Fix |
|---|---|
| netresearch/agent-harness-skill | [#65](https://github.com/netresearch/agent-harness-skill/pull/65) — two bullets for files that never existed removed |
[...]

PR #268, opened 2026-08-29
fix(validate): a bare directory name is not a dangling path

CybotTM:
The check from #266 flagged dxp-frontend-license for the token `assets/` in a sentence about a project's own asset directory. A bare directory name is prose about a kind of place, not a file the agent is told to open, so `scripts/`, `references/`, `assets/` and `evals/` on their own are now skipped. The test fixture carries one; the case fails without the skip (verified by stashing only the validator change). Refs #267.


CybotTM:
Self-review: 89b891b3712d346511a80339324dde1f0014d028

PR #274, opened 2026-08-29
docs(skill-repo): a shell variable in the skill text defeats the allowed-tools rule

CybotTM:
Third trap for the `allowed-tools` section, found while converting the five remaining fleet repositories (netresearch/matrix-skill#127).

`matrix-communication` opened its command block with

```bash
S=${CLAUDE_SKILL_DIR}/scripts
set +H && uv run $S/matrix-send-e2ee.py ROOM "message"
```

That runs correctly — the shell expands `$S`. But permission matching sees the command **as issued**, the literal `$S/…`, so a rule naming `${CLAUDE_SKILL_DIR}/scripts/*` matches none of it. All 20 documented calls in that skill were affected: converting the frontmatter alone would have produced a prompt on every single one, and the change would have looked done while making the skill unusable in default mode.

The section already covered the prefix trap (`bash x.sh` vs `./x.sh`) and the separate-tools trap. This is the same family — the rule has to match the command that is actually issued — but it is invisible in a way the other two are not: the path *looks* right in the source.

Fix is to write the path out. The line budget affords it: that skill sits at 61 of 500 body lines.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01JYQciiXoiApXBfcJrFMnA9)_


CybotTM:
Self-review: ad8824aba1eb5c8685fa5f901be4569a4a45e1b2

PR #276, opened 2026-08-29
docs(authoring-ci-gotchas): parse a structured value, do not pattern-match it

CybotTM:
Adds section 8 to `authoring-ci-gotchas.md`, from the review history of #275.

## What happened

The `allowed-tools` check in `validate-skill.sh` started as one anchored `grep`. It took **four review rounds**, and each found another legal spelling of the same value:

| Round | Spelling that slipped past |
|---|---|
| 1 | folded scalar `>-` and the YAML list form — the check read only the key line |
| 2 | a blank line inside the value, which ended collection; `Bash,Read` and `"Bash"`, where the boundary was whitespace-only |
| 3 | a YAML comment mentioning what it does *not* grant, matched as if it did |
| 4 | flow list `[Bash]`, where `]` was not a delimiter |

Widening the pattern buys one shape per round. The signal that the *approach* was wrong rather than incomplete is the repetition itself: the value never changed, only its spelling.

## What the section records

The four steps that ended it — collect the whole value ending only at a new top-level key, drop comments, split at parenthesis depth 0, match each entry anchored — plus two things that only showed up in practice:

- anchoring the comment strip on `[^)]*$` fails on a comment containing a parenthesis, which is exactly what a comment about `Bash(python3:*)` is;
- a naive split broke two *existing* passing tests, because `Bash(git:*,make:*)` has internal commas and `Bash(bash ${CLAUDE_SKILL_DIR}/scripts/*)` has a space. The suite caught it, so the section names the shape cases to write before widening anything.

## Verification

```
validate-skill.sh: 0 errors
10/10 repository tests pass
pre-commit run --files <changed>: passes
```

Documentation only.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01JYQciiXoiApXBfcJrFMnA9)_

### DIFF
--- a/skills/skill-repo/references/authoring-ci-gotchas.md
+++ b/skills/skill-repo/references/authoring-ci-gotchas.md
@@ -9,6 +9,7 @@
 - Installing the Claude Code CLI with `--ignore-scripts`
 - Generated YAML: exactly one trailing newline
 - `MD010` breaks copy-pasted Makefile snippets — exempt the fence, don't fake the tab
+- A validator over a structured value parses it; it does not pattern-match it
 
 Process learnings from a cross-session retrospective (2026-06-27). Companion to
 [`skill-quality.md`](skill-quality.md) (SKILL.md sizing) and the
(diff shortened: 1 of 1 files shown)


CybotTM:
Self-review: 1692d400bccb8fa50998c0cf012970033f8d220e

PR #280, opened 2026-08-31
docs(evals): say that a trigger query must not be copied from a benchmark

CybotTM:
Rule 7 told authors to take trigger-query phrasings from requests real users actually made. Read literally, that produced the defect it exists to prevent: the first trigger file written under it took its queries verbatim from [agent-system-evals](https://github.com/netresearch/agent-system-evals), named that repository's case identifiers, and recorded the routing rate each case had measured — including the rate for the case that was measuring the skill at the time. The benchmark's contamination check raised nineteen hits the moment a fleet resolved the skill; the arm had to be rebuilt and the comparison re-run ([typo3-conformance-skill#120](https://github.com/netresearch/typo3-conformance-skill/pull/120), [agent-system-evals#37](https://github.com/netresearch/agent-system-evals/pull/37)).

The line is between subject and sentences. A trigger eval for a conformance skill is about version declarations, and so is the case that measures it — sharing that subject is unavoidable and carries nothing. Sharing the case's sentences hands the agent the prompt it will be given, and a case identifier hands it the answer key.

The identifier half is now mechanical. `validate-evals.sh` fails on a case identifier anywhere in `eval_queries.json`. The pattern is uppercase words followed by a zero-padded three-digit ordinal — deliberately not four digits, so `CVE-2024-1234` and `OWASP-A01-2021`, which a security skill may legitimately quote, do not trip it. Exercised in both directions on a real trigger file and against both of those.

The sentence half stays prose, because this repository cannot see the benchmark's instructions to diff against them; the rule says to check it, and how.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_012XjwqvXjw8sA67NoZcoSw7)_

### DIFF
--- a/skills/skill-repo/references/materialization-contract.md
+++ b/skills/skill-repo/references/materialization-contract.md
@@ -278,6 +278,24 @@ format [agentskills.io documents](https://agentskills.io/skill-creation/optimizi
 requests real users actually made rather than from the description they are
 meant to test.
 
+**Never copy them from a benchmark that measures the skill.** This advice, read
+literally, produced exactly that: the first trigger file written under it took
+its queries verbatim from `agent-system-evals`, named that repository's case
+identifiers, and recorded the routing rate each case had measured — including
+the rate for the case that was measuring the skill at the time. The benchmark's
+contamination check raised nineteen hits the first moment a fleet resolved the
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review: a319d26027d604e3629d3dea9043cce4e1b170b7

PR #295, opened 2026-09-09
fix(roll-changelog): take the compare base from the headings

### LINKED ISSUE #281: roll-changelog.py: the new release's compare link inherits the stale [Unreleased] base
When a release is rolled, the `[X.Y.Z]:` reference definition is written with the compare base taken from the *previous* `[Unreleased]` link rather than from the previous released version. Where an earlier release already left `[Unreleased]` pointing at an older tag, the new range spans several releases.

Observed in the 2026-09-03 fleet sweep, in both repos that carried a keep-a-changelog link section:

- `matrix-skill`: `[3.1.3]: …/compare/v3.1.1...v3.1.3` — should be `v3.1.2...v3.1.3`. The `[3.1.2]:` definition was also never written, which is how the stale base survived.
- `php-modernization-skill`: `[1.22.3]: …/compare/v1.21.0...v1.22.3` — should be `v1.22.2...v1.22.3`.

Both were corrected by hand in the release PRs (netresearch/matrix-skill#132, netresearch/php-modernization-skill#106), so no release shipped with a wrong range — but the next sweep reproduces it.

Expected: the compare base is the newest released heading in the file (the one the roll just pushed down), independent of what `[Unreleased]` pointed at. A missing reference definition for that previous version should be written rather than silently skipped, since its absence is what lets the stale base persist across releases.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_016Dwq4QYJGiaEeUjUyyUKup)_

CybotTM:
Closes #281.

The new release's `[X.Y.Z]:` reference definition took its compare base from the previous `[Unreleased]` link. Where an earlier release had left `[Unreleased]` pointing at an older tag, the new range spanned several releases:

- `matrix-skill`: `[3.1.3]: …/compare/v3.1.1...v3.1.3` — should be `v3.1.2...v3.1.3`
- `php-modernization-skill`: `[1.22.3]: …/compare/v1.21.0...v1.22.3` — should be `v1.22.2...v1.22.3`

Both were corrected by hand in their release PRs, so nothing shipped wrong — but every sweep reproduced it.

## What changed

The base now comes from the **released headings**: the heading directly below the one the roll just wrote. That is independent of whatever `[Unreleased]` pointed at, which is the property the issue asked for.

The second half matters as much. A previous release with no definition of its own is what let the stale base survive across releases — `matrix-skill` never had a `[3.1.2]` line, so nothing ever corrected the drift. When the roll can see the version below it, that definition is now backfilled instead of silently skipped.

Tag style is read from the existing definitions rather than assumed, so a repo using unprefixed tags keeps them unprefixed on both ends of the range.

## Tests, and why both groups exist

Both are verified as guards, each against a mutation the other does not catch — so neither is decoration:

| Mutation | What reddens |
|---|---|
| `prev_tag = old` (inherit the `[Unreleased]` base again) | 3 of the stale-base cases |
| `styled_tag` always adds a `v` | only the bare-tag case |

The stale-base fixture reproduces `matrix-skill` exactly: `[Unreleased]` still at `v3.1.1`, no `[3.1.2]` definition, rolling `3.1.3`. One case asserts the correct range is present, a second asserts the wrong one is *absent* — the pair matters, because a definition can be right and a stale sibling still linger.

The pre-existing `refs.md` case passed either way, since its `[Unreleased]` base happened to equal the previous heading. That is why the bug survived having tests.

`ruff format` and `ruff check` clean. The formatter rewrote the function after the first mutation run, so the mutation was re-run against the final file rather than trusting the earlier result.

## Round 2 — found by reading this diff, not by a bot

Both bot reviewers were unavailable (Copilot out of quota account-wide, CodeRabbit rate-limited and it refused an explicit `@coderabbitai review` with "Action not completed"), so the diff was read by hand. That turned up a defect in this PR's own code, fixed in a follow-up commit.

`have_prev` decided whether the previous version already had a reference definition by matching the heading label verbatim. Headings and definitions spell the tag independently, so a file heading `## [v3.1.2]` while defining `[3.1.2]:` was read as missing the definition and got a **second** one under the other spelling. The comparison now drops the `v` prefix on both sides — the same normalization `styled_tag` already applies.

Covered by a fixture with exactly that mismatch. Reverting to the raw comparison reddens it with `expected '1', got '2'`.

Rebased onto `main` after #294 merged; suite, `ruff format`, `ruff check` and `shellcheck` re-run on the rebased tree.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01L94yEWSBvuUkLqLpVbMqRh)_

### DIFF
--- a/skills/skill-repo/scripts/roll-changelog.py
+++ b/skills/skill-repo/scripts/roll-changelog.py
@@ -289,6 +289,31 @@ def main() -> None:
         print(defs_note)
 
 
+def released_labels(lines):
+    """Tag labels of every released heading, document order, fences skipped.
+
+    The compare base for a new release is the heading directly below it, not
+    whatever [Unreleased] happened to point at — see update_link_reference_
+    definitions.
+    """
+    labels = []
+    for _, line, in_fence in scan(lines):
+        if in_fence:
+            continue
+        for rx in (LINKED_RE, BRACKETED_RE, BARE_PAREN_RE):
+            m = rx.match(line)
+            if m:
+                labels.append(m.group("tag"))
+                break
+    return labels
+
+
+def styled_tag(label, v_prefixed):
+    """Render `label` in the tag style the link definitions already use."""
+    bare = label.removeprefix("v")
+    return f"v{bare}" if v_prefixed else bare
+
+
 def update_link_reference_definitions(new_lines, new_heading, version):
     """Keep-a-changelog ref-style links: `[Unreleased]: …/compare/vOLD...HEAD`
     at the bottom, one definition per released heading. After a roll the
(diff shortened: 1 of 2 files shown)


CybotTM:
@coderabbitai review

CybotTM:
Self-review: b05e67f50429b590a3fe3e08882b0d34881a0610

PR #296, opened 2026-09-09
feat(validate): check that README setup instructions are runnable

### LINKED ISSUE #292: No gate validates README setup instructions
## Problem

`netresearch/retro-skill#90` was an outside user reporting that our documented setup steps do not work. The wrong text had been in that README since it was written, and nothing in CI could have caught it.

The audit that followed found the same class of defect spread across the fleet:

| Defect | Repos |
|---|---|
| `marketplace add` pointing at the skill's own repo (fails outright) | 1 (retro-skill) |
| Pointing at the wrong catalog (plugin not in it) | 1 (coding-ai/dxp-frontend-skill) |
| Missing the `/plugin install` line entirely | 28 |
| Install line without the `@marketplace` suffix | 1 |

All are now fixed, but only because a user complained.

## Why CI missed it

`check-template-drift.yml` governs the `.github/` tree only. `README.md` is outside its scope, so a README may say anything — including instructions that fail on the first command — and every gate stays green. `validate-skill.sh` does not look at the install block either.

## Proposal

Add a check that asserts, for any repo whose README mentions `/plugin marketplace add`:

1. The `marketplace add` target is a catalog repo, never the repo itself. A repo shipping `.claude-plugin/plugin.json` and no `.claude-plugin/marketplace.json` can never be a valid target.
2. A `/plugin install <name>@<marketplace>` line follows it, and `<name>` matches the repo's entry in the catalog it names — not the repo slug. Three plugins in the fleet already differ from their repo name (`jira-skill` → `jira-integration`, `matrix-skill` → `matrix-communication`, `php-ast-edit-skill` → `php-structured-edit`), so deriving the name from the slug is itself the bug.
3. The named catalog actually contains that plugin. This is what would have caught `dxp-frontend-skill`, which pointed at the public catalog for a plugin published only in the internal one.

Checks 1 and 2 are local. Check 3 needs the catalog fetched at CI time, so it may fit better as a scheduled job in the marketplace repo, which already knows every entry and its source repo.

## Related

- netresearch/retro-skill#90 — the report
- netresearch/skill-repo-skill#291 — the canonical template fix
- netresearch/retro-skill#91 — the one genuinely broken repo

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01L94yEWSBvuUkLqLpVbMqRh)_

CybotTM:
Closes #292.

netresearch/retro-skill#90 was an outside user reporting that our documented setup steps fail on the first command. The wrong text had been in that README since it was written, and nothing in CI could have caught it: `check-template-drift` governs the `.github/` tree only, and `validate-skill.sh` never looked at the install block.

## What is checked

Only when the README documents a `/plugin marketplace add`:

| Assertion | Level | The defect it catches |
|---|---|---|
| The add target is not this repo | error | `marketplace add` resolves `.claude-plugin/marketplace.json` in the **target**, so aiming it at a repo shipping `plugin.json` can never work — the exact #90 failure |
| An install line follows | warning | An add line alone registers a marketplace and installs nothing; 28 repos documented exactly that |
| The part after `@` is a marketplace name, not `owner/repo` | error | The second, independent defect in the reported snippet |

**Only fenced lines count.** A README may quote the wrong form deliberately to warn against it — this repo's own does — and a version that reads the whole file reports that warning as a defect.

## Not checked, deliberately

Whether the named catalog actually lists this plugin. That needs the catalog fetched at validation time, which does not belong in an offline validator; the issue's scheduled-job option is the better home, and the omission is marked in the code so the next reader does not assume it is covered.

## Measured before adding, so it turns nothing red

This check runs fleet-wide, so the question is not whether it is correct but what it does to 55 repositories today:

- **39** document an add line → **38 clean, 0 errors**
- the single warning is `pagerangers-skill`, which is archived
- the other **16** document none, and the check stays silent

The first run of that measurement was against stale local refs and reported 25 warnings and 1 error; re-run after fetching, it is the table above. Worth stating, because the stale answer looked plausible.

## Tests

Five cases, each verified against a mutation rather than assumed to cover something:

| Mutation | What reddens |
|---|---|
| Check disabled entirely | 4 of 5 |
| Check made fence-blind | the prose case only |

The prose fixture had to be rewritten to earn that. Its first version put the wrong line behind prose text (`> **Do not** run …`), where the line-start anchor already excluded it — so it passed the fence-blind mutation and proved nothing. It now places the wrong line at column 0 outside a fence, which only the fence tracking can exclude.

`shellcheck -x` clean; the validator passes against this repo with 0 errors and 0 warnings.

## Round 2 — three defects found by reading this diff

Both bot reviewers were unavailable (Copilot out of quota account-wide, CodeRabbit rate-limited and it refused an explicit `@coderabbitai review` with "Action not completed"), so the diff was read by hand. All three defects below would have run in **every** fleet repository:

- The self-reference test substring-matched the whole line. A repo `foo` documenting `add netresearch/foo-catalog` matched the slug `netresearch/foo` and would have failed CI on a **correct** README. It now compares the argument after `add`, exactly.
- The slug hardcoded `netresearch/` and took a `basename`, so a GitLab remote under another group resolved to a name that exists nowhere. Owner and repo are now parsed from the remote URL, scp-style form included.
- The install target was cut with `${line#*install }`, which yields an empty string when the line has two spaces after `install` — and then reports the wrong diagnostic. Both targets are now parsed as fields.

The fleet measurement was **re-run** after these changes, because the slug derivation moved and the earlier number no longer described the new code: 39 repos document an add line, 38 clean, 0 errors, one warning (`pagerangers-skill`, archived).

Both new tests are verified against the defect they cover — re-injecting the substring match and the prefix strip reddens exactly those two.

Rebased onto `main` after #294 merged; tests, self-validation and `shellcheck` re-run on the rebased tree.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01L94yEWSBvuUkLqLpVbMqRh)_

### DIFF
--- a/skills/skill-repo/scripts/validate-skill.sh
+++ b/skills/skill-repo/scripts/validate-skill.sh
@@ -961,6 +961,59 @@ if [[ -f "$REPO_DIR/README.md" ]]; then
             fi
         done
     fi
+
+    # --- Install instructions must be runnable -------------------------------
+    # A README documented steps that fail on the first command for as long as it
+    # existed, and no gate could see it: an outside user reported it
+    # (netresearch/retro-skill#90). Only FENCED command lines are considered — a
+    # README may legitimately quote the wrong form in prose to warn against it,
+    # and this one does.
+    README_CMDS=$(awk '/^ {0,3}(```|~~~)/ { infence = !infence; next } infence' \
+        "$REPO_DIR/README.md")
+    MP_ADD=$(grep -E '^[[:space:]]*/plugin[[:space:]]+marketplace[[:space:]]+add[[:space:]]' \
+        <<< "$README_CMDS" || true)
+    if [[ -n "$MP_ADD" ]]; then
+        SELF_SLUG=""
+        if [[ -n "${GITHUB_REPOSITORY:-}" ]]; then
+            SELF_SLUG="$GITHUB_REPOSITORY"
+        elif git -C "$REPO_DIR" remote get-url origin &> /dev/null; then
+            SELF_SLUG="netresearch/$(basename "$(git -C "$REPO_DIR" remote get-url origin 2>/dev/null)" .git)"
+        fi
+        # `marketplace add` resolves .claude-plugin/marketplace.json in the
+        # TARGET repo. Pointing it at a repo that ships only plugin.json fails
+        # with "Marketplace file not found" — provable locally when the target
+        # is this repo.
+        if [[ -n "$SELF_SLUG" && ! -f "$REPO_DIR/.claude-plugin/marketplace.json" ]] \
+            && grep -qF -- "$SELF_SLUG" <<< "$MP_ADD"; then
+            error "README.md: '/plugin marketplace add $SELF_SLUG' cannot work — the target must be a catalog repo shipping .claude-plugin/marketplace.json, and this repo ships .claude-plugin/plugin.json. Point it at netresearch/claude-code-marketplace (its entry names this repo as the source, so the code still comes from here)."
+        else
+            success "README.md marketplace-add target is not this repo"
+        fi
+        # The add line alone registers a marketplace and installs nothing.
+        MP_INSTALL=$(grep -E '^[[:space:]]*/plugin[[:space:]]+install[[:space:]]' \
+            <<< "$README_CMDS" || true)
+        if [[ -z "$MP_INSTALL" ]]; then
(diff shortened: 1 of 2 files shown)


CybotTM:
@coderabbitai review

CybotTM:
Self-review: 0754a3c00ea0c61da5c570f2f19a7902fa18712d

Issue #218, opened 2026-08-13
Reusable fleet-release driver script — stop re-implementing release-discipline per sweep


CybotTM:
The 2026-08-13 sweep (23 releases across GitHub + git.netresearch.de) hand-wrote five one-off driver scripts: remote-first survey, bump-PR fan-out per host, merge/tag/verify gates, and a CI-adoption completion driver. Every sweep re-implements the prose in references/release-discipline.md and re-invents the same bugs — this sweep alone hit the brace-group exit killing a batch, jq -n pretty-printing the survey JSONL, and jq // passing empty strings (all now documented in #217).

Proposal: ship a maintained scripts/fleet-release.sh (or a small set) that mechanizes the documented flow: survey (GitHub + GitLab call triples, dedupe by resolved full_name, CI-only-delta filter, composer-version-field preflight), manifest emission for approval, then per-repo execution in subshell isolation — bump via bump-version.sh, changelog roll (5 heading shapes), push + PR/MR, completion-first check gating, merge, parity check from origin/main, signed tag on the remote tip, Release-object verification, cleanup — resumable, with a per-repo log and a log-count-equals-repo-count invariant.

Prior art: the five session scripts live in the session job dir and are summarized in the PR above; the gotchas they encode are all in release-discipline.md. Implementation is a separate mandate — this issue is the design record.

CybotTM:
Prior art preserved: the eight one-off driver scripts from the 2026-08-13 sweep are archived at https://gist.github.com/CybotTM/cdaa925fbe1140fe8a98c2779484d9b3 (the session job dir mentioned in the issue body is ephemeral and has been cleaned). They carry the bugs-now-documented too — the phase_c brace-group exit is the live specimen of the hazard #217 documents.

PR #297, opened 2026-09-09
feat(fleet-release): carry operator-defined trailers on bump commits

CybotTM:
A fleet whose policy requires agent or tool disclosure on every commit cannot use this driver today. `fr_commit_allowlisted` writes `git commit -S --signoff` with a generated subject and offers no way to add anything to it, so a sweep either drops the disclosure or hand-writes a driver — the outcome #218 exists to prevent.

`FR_COMMIT_TRAILERS` holds newline-separated `Key: value` lines; each becomes a `git commit --trailer`. The driver interprets none of them, so which keys a fleet uses stays an operator decision rather than this script's knowledge — `Assisted-by:`, `Agent-Session:` and vendor-neutral variants all work without a code change. Requires git >= 2.32.

The trailers are computed once and passed to **both** commit attempts. The retry after a reformatting hook is a second `git commit`, and it is the one place they could silently be dropped — for exactly the repos whose hooks rewrite the manifest `sync-plugin-manifest.sh` emits.

Unset, the commit is byte-identical to before.

## Tests

Nine checks in `tests/fleet-release.sh`: the three trailers landing, the blank-line skip, coexistence with `--signoff`, an untouched subject, survival of the reformatting-hook retry, and the unset opt-out.

Verified by mutation rather than by passing on a clean tree: removing the `"${trailer_args[@]}"` expansion reddens five of the nine (the two that stay green assert `--signoff` and the subject, which the mutation does not touch, and the unset case is the opt-out). Restoring makes the full suite pass.

## set -u safety

The array is expanded as `${trailer_args[@]+"${trailer_args[@]}"}`. Expanding an **empty** array as `"${a[@]}"` under `set -u` is an error before bash 4.4, and `fr_preflight` admits bash >= 4 — so the plain form would abort every bump on those shells, and only for operators who did *not* set the variable, which is the default path. Found while reviewing the diff; not reproducible on bash 5.2.

## Documentation

`FR_COMMIT_TRAILERS` is documented in the driver's `--help` header and in `references/release-discipline.md`, in the fleet-sweep section that tells a sweep to extend the driver rather than hand-write one.

Signed-off-by: Sebastian Mendel <info@sebastianmendel.de>
Assisted-by: claude-code:claude-opus-5
Agent-Session: https://claude.ai/code/session_01Utz6KTyv1TNKMvXpqyNSv5

### DIFF
--- a/skills/skill-repo/references/release-discipline.md
+++ b/skills/skill-repo/references/release-discipline.md
@@ -41,6 +41,14 @@ shared release CI first), and non-default-branch releases — the
 something the driver lacks, extend it in a PR; a session-local driver script
 is how the next incident starts.
 
+**Where a policy requires agent or tool disclosure on every commit, set
+`FR_COMMIT_TRAILERS`** — newline-separated `Key: value` lines that the bump
+commit carries as git trailers, alongside the `--signoff` it already writes.
+The driver appends them verbatim and interprets none of them, so the keys stay
+the operator's decision (`Assisted-by:`, `Agent-Session:`, … ). It applies to
+both commit attempts, including the retry after a reformatting hook — the one
+place a second `git commit` could otherwise drop them. Unset, nothing changes.
+
 **Budget the API before a sweep: the survey costs ~10 calls per repo, and
 each GitHub quota pool (REST 5,000/h, GraphQL 5,000 points/h — separate
 pools) is shared across every tool, watcher and agent in the session.** A
--- a/skills/skill-repo/scripts/fleet-release-common.sh
+++ b/skills/skill-repo/scripts/fleet-release-common.sh
@@ -799,6 +799,26 @@ fr_stage_allowlisted() { # REPO WT — stage every pending path, refusing anythi
     git -C "$wt" add -- "${paths[@]}"
 }
 
+fr_trailer_args() { # ARRAY_NAME — fill with --trailer args from FR_COMMIT_TRAILERS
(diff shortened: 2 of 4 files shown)


CybotTM:
Copilot is out of review quota for the month. The quota is account-wide, so no other pull request can obtain one either, and `pr-status.sh` reports the state as unsatisfiable rather than pending. Reviewed the diff myself instead.

What the review covered: the trailer lines are appended verbatim and never parsed, so the driver stays agent-neutral; `--trailer` appends into the message's own trailer block and composes with the existing `--signoff` rather than competing with it; the arguments are computed once and passed to both commit attempts, since the retry after a reformatting hook is a second `git commit` and the one place they could silently vanish; and the nine tests were verified by mutation — removing the argument expansion reddens five of them — rather than by passing on a clean tree.

It found one defect, fixed in bc6fd99a: expanding an empty array as `"${a[@]}"` under `set -u` is an error before bash 4.4, and `fr_preflight` admits bash >= 4. The plain form would have aborted every bump on those shells, and only for operators who did not set the variable — the default path. Now `${trailer_args[@]+"${trailer_args[@]}"}`.

Self-review: bc6fd99aff8b

Signed-off-by: Sebastian Mendel <info@sebastianmendel.de>
Assisted-by: claude-code:claude-opus-5
Agent-Session: https://claude.ai/code/session_01Utz6KTyv1TNKMvXpqyNSv5

PR #310, opened 2026-09-16
fix(ci): pin typing_extensions in the hash-locked pip-audit env

CybotTM:
Merging this makes the shared Dependency Audit job installable again on the runner's default interpreter; it currently fails for every fleet repo that runs it. No skill content changes.

## What was wrong

`cyclonedx-python-lib` declares `typing_extensions<5.0,>=4.6 ; python_version < "3.13"`. `pip-audit.txt` was compiled on an interpreter at or above 3.13, where that marker excludes the package, so the entry was never written. The audit job installs with `--require-hashes` using the runner's default `python3` (3.12), which does need it, and pip rejects the whole file rather than just that line:

```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    typing_extensions<5.0,>=4.6 ... (from cyclonedx-python-lib==11.11.0)
```

This is shared CI, so it is not one repo's problem — it surfaced as a red `Dependency Audit` on [netresearch/jira-skill](https://github.com/netresearch/jira-skill).

## Fix

Recompiled with an explicit `--python-version 3.12` floor, which keeps the marker in scope. That adds `typing-extensions==4.16.0 ; python_full_version < '3.13'`; the rest of the diff is patch-level drift from recompiling a six-week-old lock.

The compile-command comment in `ci-python.yml` and `dependency-audit.yml` now names the floor. That comment is the root cause: it says `uv pip compile --universal --generate-hashes`, and following it on a newer machine silently reproduces the defect.

## Testing

Controlled before/after on the interpreter the runner actually uses (Python 3.12.14), same command, only the lock file changed:

- old file: exit 1, the exact error above
- new file: exit 0
- new file on Python 3.14 (where the marker excludes the package): exit 0

`bandit.txt` was checked for the same defect class and is unaffected: same eight packages, and its one marker-conditional entry (`colorama ; sys_platform == 'win32'`) is already pinned.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01UibkvodsLSqL2cR8ZDZX7R)_


CybotTM:
Self-review: ca894c4dc94d9cd3899d141f51b60c628035c4e7

PR #324, opened 2026-09-17
docs(authoring-ci-gotchas): renovate regex managers, SAST interpreters, script modes

CybotTM:
After this merges, `authoring-ci-gotchas.md` carries three failure modes that cost a round each in one session and were in none of its ten existing sections. Two of them are silent by construction: the tool reports success while doing less than you think.

## 11 · A Renovate regex manager fails silently in both directions

The `customManagers` entry that pins `uvx ruff@…` in a workflow has two failure modes, and **neither produces output**, because "no dependency found here" and "nothing to update here" look identical from outside.

*It stopped matching.* The pattern was written against `uvx <name>@<version>`; sixteen minutes later another commit hardened the pin to `uvx --no-build ruff@0.16.0`. From then on it matched nothing, and the pin sat unmoved for seven weeks while ruff released eight versions. The repository looked current the whole time.

*It matched the wrong token.* Widening the prefix to "any word may stand here" fixed that and broke the other direction: under a `datasource=pypi` comment, `uvx --from git+https://github.com/o/r@2843b87 tool` yields `2843b87` as the PyPI version of the package the comment names.

The section gives the one-call check against the real file (a count, which is the only thing that tells the two silent states apart), the reason to shape the pinned token as a package name rather than as `\S+`, and why a flag allow-list is not the fix — **RE2 has no lookahead**, so such a list would have to exclude the value-taking flags from its own generic branch to be safe. Plus the requirement that a test pin the shapes which must yield *nothing*; a case list asserting only what must match cannot catch the second failure at all.

## 12 · A SAST job's interpreter bounds what it scans, and says nothing

`python3 -m venv` in a CI job inherits the runner image's interpreter — 3.12.3 on `ubuntu-latest` today, which for most callers is the **oldest** version their matrix tests. Bandit parses with the `ast` of the interpreter it runs on, and a file it cannot parse is skipped with a warning while the run still exits 0. The gate reports clean rather than reporting that it scanned nothing.

Measured with bandit 1.9.4 on one file holding a PEP 696 type-parameter default and a shell-injection finding under it:

| Interpreter | Result |
|---|---|
| 3.12.14 | `Files skipped (1): syntax error while parsing AST` · exit 0 · no `B602` |
| 3.14.7 | `B602 subprocess call with shell=True` · exit 1 |

The section derives the interpreter from what the caller tests and asserts the venv landed on it, and names the two details that each cost a round: `uv venv --seed` is what provides the pip that reads a hash-locked file under `--require-hashes`, and `gsub` before `tonumber` because a caller may test a free-threaded build (`3.13t`) where a bare `tonumber` aborts the job — with the matching note that `python -V` then answers `Python 3.13.x`, so the assertion compares the numeric prefix.

## 13 · A new script is committed `100755`

ruff's `EXE001` fails the build on a shebang committed `100644`, and a local ruff run does not reproduce it. `validate-skill.sh` already catches this with the exact fix, but only after a push — this cost two separate CI rounds in one session, on a maintainer's PR and a contributor's, both on freshly added test scripts. One line at creation time removes both.

## Evals

Two TDD stubs, since `evals.json` had nothing in either area: one asserts the verification-after-a-flag-change answer for §11, one the interpreter answer for §12. They are stubs in the sense that they pin the *presence* of the guidance, not its wording.

## Scope

One reference file and `evals.json`. No script, no SKILL.md, no workflow, no version bump. The ten existing sections are untouched; the Contents list gains three lines.

## Testing done

`markdownlint-cli2 0.22.1` clean · `validate-skill.sh .` → 0 errors, 0 warnings · `validate-evals.sh` → 49 passed, 0 failed, and both new evals named in the pass list. The pre-commit `ruff-format` hook reformatted the Python snippet inside the §12 fence; that reformat is in the commit rather than fought.

https://claude.ai/code/session_013K8mwWMSxXGZnwu93Di3p9


CybotTM:
**No bot review was obtainable on this head, so the diff was reviewed by hand. Recording what that covered.**

Both reviewers are walled: Copilot reports `copilot_quota_exhausted: true` with `next.reason: bot-review-unsatisfiable`, and CodeRabbit answered `⚠️ Action not completed` / `Review rate limited.` Neither changes by waiting.

### What the hand review looked at

`git diff origin/main...25a7539` — one reference file and `evals.json`, 157 insertions, 4 deletions, two commits.

It found three defects, all fixed in `25a7539` before this note:

- **Section 12 told the reader to assert the venv landed on the chosen interpreter, and then showed a snippet that only prints `python -V`.** Copied as written it asserts nothing — the same failure the section is about, one level up. The comparison is in the snippet now, on the numeric prefix so a free-threaded request (`3.13t`, where `python -V` answers `Python 3.13.x`) does not fail it.
- **Section 11's probe indexed `customManagers[0].matchStrings[0]`**, so a repository with more than one manager silently measures the wrong pattern while printing a plausible count. It iterates now. Run verbatim against this repository it prints `2 ['0.16.8', '0.16.8']` — both ruff pins, the state the manager was repaired into earlier today.
- The file's dateline attributed all of it to the 2026-06-27 retrospective while §§11–13 come from 2026-09-17.

### What was run

`markdownlint-cli2 0.22.1` clean · `validate-skill.sh .` → 0 errors, 0 warnings · `validate-evals.sh` → 49 passed, 0 failed, both new evals named in the pass list · the §11 probe run verbatim against this repository · CI 24 green, 0 red, 1 skipped, 0 unresolved threads.

### What this review cannot cover

The §12 bandit guidance is not exercised here: no workflow in this repository calls `ci-python.yml` with `run-bandit: true`, so the path has no CI coverage. Its first real run is `netresearch/jira-skill`'s next scan. The two eval stubs pin the *presence* of the guidance, not its wording, and no eval runner executes them in this PR.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_013K8mwWMSxXGZnwu93Di3p9)_

CybotTM:
Self-review: 25a7539de9965e984960a697b0bfea30e9fb76e8

PR #327, opened 2026-09-18
docs(architecture): what a reference reaches, measured

CybotTM:
Merging this replaces one asserted cost in the three-surfaces table with a measured one, and adds the section it points at.

The table said a gap in `references/` costs "nothing — until it is needed". That is right for a lookup table. It is wrong for a rule that prevents a common mistake, and the gap is now countable.

**`OFR-TYPO3-UPGRADE-001`, 14 September 2026, Haiku 4.5, six trials on one fleet.** This is a case where routing works — `skill_invoked` is 6 of 6, so every trial opened the skill and read the body. `Read` calls against `references/` per trial:

| trial | reference files read | outcome |
|---|---|---|
| 1 | 0 | passed both legs |
| 2 | **0** | passed v14.3, lost v13.4 |
| 3 | 3 | failed both legs |
| 4 | 2 | passed both legs |
| 5 | 0 | passed both legs |
| 6 | 0 | passed both legs |

Trial 2 lost the leg that had been working to a rule that had been sitting in `references/upgrade-v13-to-v14.md` for nine days, written from an earlier occurrence of the same failure — and it was one of the four that opened nothing. `SKILL.md` names the problem at exactly the right step and then hands off: "`createMock` on one of them cannot be repaired by swapping the name — see `references/upgrade-v13-to-v14.md`". The agent was told a rule exists and not what it says.

**What these six trials do not say:** whether reading helps. Trial 3 read three files and failed both legs; trials 5 and 6 read none and passed. Six trials cannot separate that and the section says so. What they do say is the frequency — a reference is opened in a minority of runs even when the body points straight at it.

So the rule that follows is about placement rather than length: a reference is for what an agent looks up once it knows it needs it; a sentence that has to land every time belongs in the body, even when the surrounding treatment stays in the reference.

The companion change is [netresearch/typo3-extension-upgrade-skill#94](https://github.com/netresearch/typo3-extension-upgrade-skill/pull/94), which moves that one sentence into step 9 and leaves the full treatment where it was.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01KfKeZEKvGvXvcf2vnVzUFJ)_


CybotTM:
CodeRabbit reviewed this head and returned no actionable comments; Copilot has no monthly quota, and this repository's ruleset asks for it. So I reviewed it myself — it is a documentation change whose whole content is six numbers, and what I checked is that each one comes from the artefacts.

The reference-read counts per trial from `verifier/trajectory.json` in each of the six job directories, where `tool_calls` sit under `steps[]`. That detail matters here: read from the top level the list comes back empty, and a per-trial count of zero everywhere would have looked like a much stronger finding than the true one. `skill_invoked` 6 of 6 from the same files, through the reading `scripts/analyze` uses. The per-trial outcomes from each `matrix-13.4.txt` and `matrix-14.3.txt`. The nine days between the rule landing (`e00db0c0`, 5 September) and the round (14 September) from the skill repository's commit dates, against `v3.12.5` published 13:42 that day and the round's first trial at 13:51 — so the round did carry the rule.

Two limits the section states and I want stated here too.

**Six trials cannot say whether reading a reference helps.** The trial that read three files failed both legs; two that read none passed. The claim is only the frequency, and the placement rule follows from the frequency rather than from an effect.

**It generalises from one case on one model.** `OFR-TYPO3-UPGRADE-001` under Haiku 4.5 is where routing works well enough to ask the question at all; a larger model with a bigger budget may open references far more often. The section says "measured at four of six runs" rather than "agents do not read references", which is the width the evidence carries.

CybotTM:
Self-review: 051fd746fe7ddf70d3ffb437d2ef8b974df65334

PR #329, opened 2026-09-18
docs(architecture): the one row now has its measurement

CybotTM:
Merging this gives the review-case row in "What gets executed, measured" the measurement its own caveat asked for. [#328](https://github.com/netresearch/skill-repo-skill/pull/328) called it "one row and not a mechanism" because the equipped arm differed from the bare one by eight skills besides the grep block. Since then the block alone was removed and measured ([netresearch/agent-system-evals#100](https://github.com/netresearch/agent-system-evals/pull/100)): cost did not rise, the arm without the block ran equivalent greps by hand from the tokens the body's numbered steps name in prose, and outcome held. The row keeps its claim — the block is the shape being run — and gains the sentence that it is not where that case's saving lives, with the record named. One paragraph, plus a rewrap of the line the previous correction left at 130 columns. `validate-skill.sh`: 0 errors.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01KfKeZEKvGvXvcf2vnVzUFJ)_


CybotTM:
CodeRabbit's result on this head is noted below; Copilot has no monthly quota and this repository's ruleset asks for it. What I checked myself — one paragraph added to one row:

The three claims in the added sentence against the record they cite (`experiments/OFR-TYPO3-EXT-001-20260918-170620.json` in `netresearch/agent-system-evals`, written up in that repository's #100): cost did not rise — $0.13/0.17/0.11 with the block, $0.09/0.10/0.12 without, p 0.200 on the declared endpoint; the arm without the block ran equivalent greps by hand — quoted from the three candidate trials' Bash arguments, `grep -r "declare(strict_types" … | wc -l` and its siblings; outcome held — `outcome_quality` 3/3 on both arms. The rewrapped line was the one the previous correction left at 130 columns; the diff is that paragraph and nothing else. `validate-skill.sh`: 0 errors.

One thing a reviewer should weigh. **The sentence says what the block is not, and stops.** The write-up behind it names a next hypothesis — that the body's numbered steps, not its greps, are what the equipped arm saves with — and this reference does not carry that, because it is one round's reading of a contrast between two other rounds, not a measurement. It goes in when it has one.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01KfKeZEKvGvXvcf2vnVzUFJ)_

CybotTM:
Self-review: 689d166f97bad3ee990224f5c4cf727b4e985be0

PR #335, opened 2026-09-20
chore(release): v2.2.0

CybotTM:
Merging this sets the version surfaces to 2.2.0 so the tag can be cut. It ships eleven commits, of which one changes what consumers get.

## Why minor, not patch

The release workflow now wraps every archive in one top-level folder ([#333](https://github.com/netresearch/skill-repo-skill/pull/333)). What a consumer unpacks therefore changes shape, and a patch release should not alter the form of a published artifact silently.

## What that fix does

Every skill and plugin archive this repository's reusable release workflow produces unpacks into `<name>/` instead of spilling its contents at the archive root. Two things were broken by the old shape:

- the OpenAI skills API takes *"a .zip that contains a single top-level folder"*, so a flat archive was refused with a bare **Invalid skill** — reported in [netresearch/german-technical-writing-skill#2](https://github.com/netresearch/german-technical-writing-skill/issues/2)
- every repo's README says *"Download the latest release and extract to your agent's skills directory"*, and a flat archive unpacked `SKILL.md` and `references/` straight into `~/.claude/skills/`

`tests/release-archive-layout.sh` extracts the packaging step out of the workflow and runs it against a fixture, so the property cannot regress silently.

## Reach

The fix reaches a consuming repository with **that repository's** next release; archives already published keep their layout. This release is what makes the fixed workflow available to them.

## The rest

Nine dependency and documentation commits: two Renovate pre-commit bumps, `softprops/action-gh-release` to v3.0.3, an architecture-doc row, and the follow-up that corrected the extraction destination in `installation-methods.md` and made the layout test assert the plugin manifest.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NLJXhUjDbihKeu1d51gwWD)_


CybotTM:
Self-review: d8727dfbc4f0dd063856336b0807e233aa3a1c55

The bot review this pull request demands is unsatisfiable (Copilot quota wall or repeated bot failures on this head). The diff on this head was reviewed by the PR author; this comment is the on-the-record attestation the merge gate reads back. It stops matching on the next push.

PR #331, opened 2026-09-18
docs(architecture): the second model has its row

CybotTM:
Merging this replaces the closing limit of "What gets executed, measured" — "one model at one budget" — with the second model's row. On the same case, the same skill body naming the same script (`typo3-extension-upgrade`, v3.11.1 through v3.12.5, every one naming `scripts/scan-deprecations.sh`), `claude-opus-5` ran the script by its printed path in 16 of 17 valid trials, on 20 August 2026 at $20–34 a trial; `claude-haiku-4-5-20251001` in 1 of 188, 31 August to 18 September, under $2 a trial. Six Opus trials of 19 August are excluded as invalid (two steps each, no cost). The paragraph says the rows above are Haiku's and why the stack is measured on that model.

Counted from the Bash arguments of every `OFR-TYPO3-UPGRADE-001` `verifier/trajectory.json` in the equipped arms (`nr`, `nr-full`, `candidate`) of `netresearch/agent-system-evals`, model read from each job's `config.json`, cost from `final_metrics.total_cost_usd`. The one Haiku trial that ran it is `nr-20260831-204420/…mNjCgCr`; the one valid Opus trial that did not is `nr-20260820-044333/…AWN69Vn`. One bullet replaced, nothing else. Longest line 373 columns, within the file's existing wrap. `validate-skill.sh`: 0 errors, 0 warnings.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01KfKeZEKvGvXvcf2vnVzUFJ)_

PR #336, opened 2026-09-20
chore(release): v2.1.6

CybotTM:
Ships the release-archive fix: every archive now unpacks into one top-level folder, which is what the OpenAI skills API requires and what the READMEs already told readers to expect.

- fix(release): wrap every release archive in one top-level folder
- fix(release): correct the extraction destination, assert the plugin manifest
- docs(architecture): the second model has its row


CybotTM:
Superseded by [v2.2.0](https://github.com/netresearch/skill-repo-skill/releases/tag/v2.2.0), which was released from [#335](https://github.com/netresearch/skill-repo-skill/pull/335) while this fleet sweep was running and covers the same range: [#331](https://github.com/netresearch/skill-repo-skill/pull/331), [#333](https://github.com/netresearch/skill-repo-skill/pull/333) and the dependency updates. Bumping to 2.1.6 now would move `plugin.json` backwards.

Issue #119, opened 2026-06-11
validate-skill.sh description check rejects valid YAML scalar styles (single-quoted, block scalar)


powli:
## Bug

The "description must start with 'Use when'" check in `scripts/validate-skill.sh` (v1.24.0, ~lines 87-93) extracts the raw `description:` line via grep/sed and strips only **double** quotes:

```bash
DESC=$(echo "$FRONTMATTER" | grep "^description:" | head -1 | sed 's/description: *//' | sed 's/^"//' | sed 's/"$//')
```

This rejects descriptions that are perfectly valid YAML and whose *parsed* value does start with "Use when":

- single-quoted scalars: `description: 'Use when …'` → fails (leading `'` not stripped)
- block scalars: `description: >-` → fails (DESC becomes `>-`)

The side effect is real: the check pushes authors toward unquoted single-line or multi-line *plain* scalars. Multi-line plain scalars in a SKILL.md description are invalid YAML and Claude Code then silently loads the skill with **empty frontmatter**, disabling auto-activation entirely — we shipped exactly that bug in `ecom-orocommerce-docker` ≤1.0.4 and found it via this validator's formatting pressure.

## Suggested fix

Parse the frontmatter instead of grepping the line, e.g.:

```bash
DESC=$(yq --front-matter=extract '.description' "$SKILL_MD")
[[ "$DESC" == "Use when"* ]] || error "…"
```

(or `sed -n '/^---$/,/^---$/p' | yq …` if yq's front-matter mode is unavailable). Then any YAML-valid quoting/scalar style passes as long as the parsed value starts with "Use when".

PR #338, opened 2026-09-20
feat(evals): require samples on an eval a PR adds or tightens

CybotTM:
Merging this makes `samples` a requirement on every eval a pull request adds or whose assertions it changes, enforced where the fleet's eval gate already runs: `validate-evals.sh` compares the `evals.json` under test against the same file at the base revision and fails the new or changed evals that carry none. A consumer repository sees one extra step in its `Eval Validation` job on pull requests — "Resolve the base copy of evals.json" — and an unchanged verdict on everything it already ships: the 485 fleet evals without samples are not retrofitted, push builds are untouched, and a repository whose PR adds no eval and edits no assertion notices nothing.

This is the enforcing half of the decision on [netresearch/retro-skill#92](https://github.com/netresearch/retro-skill/issues/92), option A. The rule half is [netresearch/retro-skill#119](https://github.com/netresearch/retro-skill/pull/119), which also found the limit this change closes: retro-skill has no `evals.json` and does not call `eval-validate.yml`, so its check can only bind the evals retro itself writes. Option B (adopting the `claude plugin eval` layout) was rejected there and is not built here.

**How the base is obtained, and where it is not.** `eval-validate.yml` checks out the synthetic merge commit at `fetch-depth: 1`, so no base is in the tree. On `pull_request` the new step fetches `github.event.pull_request.base.sha` and writes that revision's copy of the file to `$RUNNER_TEMP`; the validator reads it from `EVALS_BASE_FILE`. The comparison is therefore head against the base branch tip, not a computed merge base — a shallow checkout has no merge base to compute, and the difference cannot produce a false rejection: an eval the PR does not carry is never examined. Three honest gaps are handled explicitly rather than papered over: a file that does not exist on the base branch becomes `[]`, so every eval in it counts as new; a base commit that cannot be fetched skips enforcement and says so with a `::notice::`; and a base copy that exists but cannot be read or parsed is a hard failure, never an empty base, because "the base could not be read" and "the base had no evals" must not be the same state. On `push` there is no honest base at all — `github.event.before` is all-zeros on branch creation and meaningless after a force-push — so the requirement stays off there.

**What is exempt, and one thing that is exempt for a reason worth a decision.** An eval carrying no pattern this validator would run against a sample is never asked for samples: it *fails* samples that no assertion backs, so demanding them would be a gate nothing can pass. That covers the exclusion #119 named (an eval graded by `expectations` alone), a lone `*_contains` literal `grep -E` cannot parse — and, in practice, plain-string assertions. Measured over the 23 `evals.json` installed on one host: 518 evals, of which 296 would be rejected if they were new, 0 are rejected as unchanged, 12 already carry samples and 210 use plain-string assertions. Those 210 are the finding: `run-ab-evals.sh` *does* grade a plain-string assertion, falling back to `str(a)` as the pattern, while the samples machinery in this validator only ever looks at `{type, pattern}` objects. Making them sampleable is a change to what `samples` means for evals that already carry them, so it is deliberately not in this PR — it needs a maintainer's call, and #92 is the thread for it.

**Transition.** `eval-validate.yml` always runs the script from `.skill-repo-tools` at `ref: main`, so this repository's own PR run exercises the new workflow step against the *old* script, which ignores `EVALS_BASE_FILE` entirely — the step is inert until this lands on `main`, and the pairing cannot half-apply in either order. The base copy travels in the environment rather than as a flag for exactly that reason: the old script's argument loop is `*) EVALS_FILE="$arg"`, last wins, so a `--base <path>` would have made it validate the base file and go green by accident. What proves the script is the bash suite, which `tests-caller.yml` runs over `tests/**/*.sh`.

**Identity, and the one shape where "untouched" can be wrong.** An eval is keyed by `eval_name`, then `name`, then `id` — the same key [#119](https://github.com/netresearch/retro-skill/pull/119) uses, deliberately, so both halves of the rule name an eval the same way. For a file whose evals carry only an `id` (67 of the 518 measured, in two files: `german-technical-writing` 27/27 and `github-release` 40/40), an eval *appended* at the end shifts nothing and only the appendee is flagged, but an eval *inserted mid-list* renumbers every later id, and each shifted eval then compares against its neighbour's assertions — measured at 13 of 13 flagged for one insert into a twelve-eval file. Named evals are immune. The mitigation is to give the eval a name or to append; changing the key would break parity with #119, which is a decision for #92 rather than a patch here.

**One delta between the check and the rule prose:** this requires `samples.passing` to be a non-empty string, which is exactly #119's `_has_samples`, while the rule text (and the failure message) also asks for at least one `samples.failing`. Parity with the other half of the rule won over strictness; a `passing`-only samples block still gets the existing validation, which fails a passing sample any assertion rejects.

**Tests.** 13 new scenarios, 20 assertions, in `tests/validate-evals.sh` — the suite goes from 23 to 43 cases — both directions on every rule: a new eval without samples rejected and named, a new eval with samples accepted, a file compared against itself accepted, changed assertions without samples rejected, samples added to an existing eval accepted, a prompt edit that leaves the assertions alone accepted, a new expectations-only eval accepted, an id-only eval appended (only the appendee flagged), an unreadable and a malformed base copy each failing with their own message, a base copy ignored with a warning in multi-file discovery mode, and the same file passing with `EVALS_BASE_FILE` unset. Seven mutations were run, each reddening only its own cases and every run executing the full case count of 43: never reject (6 red), treat every eval as new (9), drop the exemption (1), compare the whole record instead of the assertions (1), keep the base applied in discovery mode (1), turn an unreadable base into an empty one (1), key an eval by its position (1). The script was restored from a copy after each and the suite is green on the restored tree.

**Gates**, enumerated from `.github/workflows/*.yml` and run with their own command strings: `validate-skill.sh .` (0), `tests/validate-skill.sh` (0), `tests/portable-manifest.sh` (0), `check-version-parity.sh` (0), `sync-plugin-manifest.sh --check` (0), `shellcheck -x -S style` over all 24 scripts with the pinned 0.11.0 (0), `uvx ruff@0.16.8 check` and `format --check` (0, 25 files), `yamllint -c .yamllint.yml .` (0), `actionlint` (0), `markdownlint-cli2 "**/*.md"` (0 issues, 21 files), every `tests/**/*.sh` (11 files, all 0) and `tests/**/*.py` (0), and `pre-commit run --files <changed>` (all passed or skipped); the working tree held exactly the five changed files afterwards. This repository's own `evals.json` compared against its copy on `main` with the new script: 49 passed, 0 failed.

**Not fixed here, named because it is adjacent:** the workflow's `Find evals.json` step `break`s at the first match, so a consumer shipping several skills gets only one of its `evals.json` base-compared, although the script itself validates all of them. That is the pre-existing #214 class and outside this mandate.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NLJXhUjDbihKeu1d51gwWD)_


CybotTM:
Self-review, because both bots are unavailable on this head: Copilot is out of review quota account-wide and CodeRabbit refused this head as rate-limited. I read the whole diff rather than relying on the green checks.

What I checked and accept: the base is resolved only on `pull_request`, because `github.event.before` is meaningless after a force-push — that is the right call and the comment says why. An unreadable base is a hard `FAIL` rather than an empty index, which is the distinction that decides whether "the base could not be read" silently demands samples from every eval. The identity key prefers `eval_name`/`name`/`id` over position, so inserting an eval mid-list does not report every later one as changed; where a file's evals carry no name the fallback is the index, and that residual case is named in the body. The exemption reads `patterns`, which is empty for any eval this validator could not run against a sample, so the rule cannot demand samples the validator would then fail.

One property a reader should know about, which I am recording rather than changing: a failed `git fetch` of the base commit prints a `::notice::` and leaves the requirement off for that run. The gate can therefore be disabled by a network blip without the job going red. The alternative — failing the job — trades a silent gap for a flaky gate, and with the requirement being new I would rather have the flake-free version first and tighten it if the notice is ever actually observed. The notice is the thing to watch for.

Verified independently of the agent that wrote this, with the validator from this branch against the 23 `evals.json` installed on this host: **296 evals would be rejected if they were new, 0 are rejected when each file is compared against itself.** That second number is the property the change lives or dies by, and it is the one I re-measured rather than read.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01NLJXhUjDbihKeu1d51gwWD)_

CybotTM:
Self-review: ae1733673d2902b32b71e1767551099887118f16

The bot review this pull request demands is unsatisfiable (Copilot quota wall or repeated bot failures on this head). The diff on this head was reviewed by the PR author; this comment is the on-the-record attestation the merge gate reads back. It stops matching on the next push.

